PR #95, opened 2024-12-27
Update 'create a toolkit' docs

EricGustin:
Tell readers to use make commands for their newly created toolkits. Also some other small improvements to the doc.


Spartee (approved):
LGTM

PR #129, opened 2025-01-25
Clarify use of OpenAI with Arcade

sdserranog:
To use arcade tools with OpenAI, you'll need both `tool_choice` and `user`. This is a small update to our docs to clarify that!
Suggestions are welcome.

https://app.clickup.com/t/86b353pkf


EricGustin (approved):
The OpenAI example feels like a JavaScipt example. Anyway to tabulate the language for the OpenAI example?

.prettierrc
```diff
@@ -1,4 +1,5 @@
 {
   "proseWrap": "never",
-  "plugins": ["prettier-plugin-tailwindcss"]
+  "plugins": ["prettier-plugin-tailwindcss"],
+  "parser": "markdown"
```
sdserranog, line 4: This should fix prettier breaking code blocks.

EricGustin (approved):

PR #138, opened 2025-01-28
Set root key when installing locally

nbarbettini:
Without this, the Engine will not run properly.


nbarbettini:
This is no longer needed. We set a default root key in `engine.env` which means the dev doesn't need to worry about this for local dev.

cc @sdreyer thanks for pushing me in this direction!

PR #139, opened 2025-01-29
Add js examples to auth providers

sdserranog:
More JavaScript examples for our docs, and fixing a misconfiguration in Prettier I added in my last commit.

### DIFF
--- a/examples/code/integrations/atlassian/custom_auth.js
+++ b/examples/code/integrations/atlassian/custom_auth.js
@@ -17,6 +17,5 @@ if (authResponse.status !== "completed") {
 // Wait for the authorization to complete
 authResponse = await client.auth.waitForCompletion(authResponse);
 
-// eslint-disable-next-line no-unused-vars, @typescript-eslint/no-unused-vars
 const token = authResponse.context.token;
 // Do something interesting with the token...
--- a/examples/code/integrations/atlassian/custom_auth.js
+++ b/examples/code/integrations/atlassian/custom_auth.js
@@ -1,0 +1,22 @@
+import { Arcade } from "@arcadeai/arcadejs";
+
+const client = new Arcade(); // Automatically finds the `ARCADE_API_KEY` env variable
+
+const userId = "user@example.com";
+
+// Start the authorization process
+let authResponse = await client.auth.start(userId, "atlassian", {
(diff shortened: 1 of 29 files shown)


examples/code/integrations/atlassian/custom_auth.js
```diff
@@ -0,0 +1,22 @@
...
+
+if (authResponse.status !== "completed") {
+  console.log("Please complete the authorization challenge in your browser:");
+  console.log(authResponse.url);
+}
+
+// Wait for the authorization to complete
+authResponse = await client.auth.waitForCompletion(authResponse);
+
+// eslint-disable-next-line no-unused-vars, @typescript-eslint/no-unused-vars
```
wdawson, line 20: We should remove these lint disable comments. Can we do something with token on a later line that we'll never show in the docs? can use the syntax:

```md
```javascript file=<rootDir>/examples/code/integrations/atlassian/custom_auth.js#L1-L22 {8-10}

```

examples/code/integrations/dropbox/custom_auth.js
```diff
@@ -0,0 +1,22 @@
+import { Arcade } from "@arcadeai/arcadejs";
+
+const client = new Arcade();
+
+const userId = "user@example.com";
+
+// Start the authorization process
+const authResponse = await client.auth.start(userId, "dropbox", {
+  scopes: ["openid", "profile", "email", "guilds.join"],
```
wdawson, line 9: these scopes are wrong (my bad!) also wrong in the python example. I think I copied from discord and never replaced. Can you update these and the python example to be these scopes instead?

- `openid`
- `sharing.read`
- `files.metadata.read`

wdawson (commented):
Couple comments, but looks great overall!

I'm looking forward to when we get the custom tool definition too 🚀

examples/code/integrations/dropbox/custom_auth.js
sdserranog: Thank you for catching this!

examples/code/integrations/atlassian/custom_auth.js
sdserranog: I excluded the examples folder from the linter

wdawson (approved):

PR #142, opened 2025-01-29
Add dashboard button to navbar

EricGustin:


src/components/SignUpButton.tsx
```diff
@@ -1,13 +1,15 @@
 import Link from 'next/link'; import { Button } from './ui/button';
```
sdserranog, line 1: We need to fix prettier, and we can also pass the variant as a prop, using variant="outline" for the button 

Screenshot 2025-01-29 at 7 24 17 PM

src/components/SignUpButton.tsx
```diff
@@ -1,13 +1,15 @@
...
 
-export const SignUpButton = () => (
+interface SignUpButtonProps { href: string; text: string; }
+
+export const SignUpButton = ({ href, text }: SignUpButtonProps) => (
```
sdserranog, line 5: We might also consider changing this name, as it is no longer just a signup button.

src/components/NavBarButton.tsx
```diff
@@ -0,0 +1,32 @@
+import Link from "next/link";
+import { Button } from "./ui/button";
+
+interface NavBarButtonProps {
+  href: string;
+  text: string;
+  variant?: string;
```
sdserranog, line 7: There’s no need to do this, but we can get Intellisense to autocomplete the types if we add it here instead of inside the component:
```suggestion
  variant?:
    | "default"
    | "link"
    | "destructive"
    | "outline"
    | "secondary"
    | "ghost";
```

sdserranog (approved):
Thank you for doing this!  :shipit:

PR #143, opened 2025-01-30
Fix vulnerabilities

sdserranog:
We're removing package-lock.json because we're using pnpm as our package manager. Also, we've addressed some vulnerabilities raised by Vanta.


wdawson (approved):
🔒

PR #144, opened 2025-01-30
Add JS examples to our slack toolkit

sdserranog:


public/examples/integrations/toolkits/slack/get_conversation_history_by_id_example_call_tool.js
```diff
@@ -0,0 +1,34 @@
...
+const USER_ID = "you@example.com";
+const TOOL_NAME = "Slack.GetConversationHistoryById";
+
+// Start the authorization process
+const authResponse = await client.tools.authorize({
+    tool_name: TOOL_NAME,
+    user_id: USER_ID,
+});
+
+if (authResponse.status !== "completed") {
```
nbarbettini, line 14: Thinking out loud: I wonder if adding a `status.isComplete()` helper would be nice here

wdawson: This is not for now, I'll file a ticket. We should think about how clients should handle errors here.

wdawson (approved):
LGTM but I know Nate is also reviewing.

nbarbettini:
LGTM, thanks for reviewing @wdawson

pages/toolkits/social-communication/slack.mdx
```diff
@@ -102,7 +102,7 @@ pip install arcade_slack
         Python: [
           "/examples/integrations/toolkits/slack/send_dm_to_user_example_call_tool.py",
         ],
-        JavaScript: ["Coming Soon"],
+        JavaScript: ["/examples/integrations/toolkits/slack/send_dm_to_user_example_call_tool.js"],
```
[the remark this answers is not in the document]
sdserranog: I want to improve the `TabbedCodeBlock` component's UI. We are also currently exposing the code in a public folder (which isn't ideal), and we can create a hook to save your language preference in local storage and use it here. I'll create a ticket for this.

PR #153, opened 2025-02-08
Update engine.env docs

EricGustin:


examples/code/home/configuration/engine/arcade.sh
```diff
@@ -1,50 +1,60 @@
-### Engine Configuration ###
-
+# Telemetry
```
EricGustin, line 1: copied from engine's distribution file

nbarbettini (approved):
LGTM

PR #158, opened 2025-02-14
Documenting new Slack tools and modifications to existing tools

byrro:
Tools added to retrieve metadata and messages from direct message conversation. Changed the name of tools referring to 'conversation by name' to 'channel by name', since the tools don't actually apply to IM or MPIM, only to public or private channels.


EricGustin (approved):
:partyparrot: emoji

PR #171, opened 2025-02-24
Add UTMs to registration urls.

sdserranog:
https://app.clickup.com/t/86b3m15j0

Add UTM parameters to the registration URL so we can track the source.

### DIFF
--- a/pages/home/api-keys.mdx
+++ b/pages/home/api-keys.mdx
@@ -8,7 +8,7 @@ import { Steps, Tabs } from "nextra/components";
 # Getting Your API Key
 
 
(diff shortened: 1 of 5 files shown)


nbarbettini:
@sdserranog We currently use Cloudflare to create the shortcut `api.arcade.dev/signup`. It is done with a Page Rule here: https://dash.cloudflare.com/8eea19ea3af73afba5044637e55cd204/arcade.dev/rules/page-rules/form

I would like to find a way to keep the `api.arcade.dev/signup` shortcut, but allow these UTM params to be appended to it. I'm not sure if that already works with the page rule as-is, can you test it?

If that doesn't work with page rules out of the box, then an alternative is creating a Worker Route with a small worker that has the biz logic to rewrite the URL as a 302 but append incoming query params to it.

nbarbettini (approved):
Thanks for getting this working @sdserranog!

PR #174, opened 2025-02-27
Fix unintentional html tag

EricGustin:
The linter thinks that I was using an html tag. Updated so that it is happy and the build passes.

### DIFF
--- a/pages/toolkits/community-toolkit-template.mdx
+++ b/pages/toolkits/community-toolkit-template.mdx
@@ -1,3 +1,3 @@
-# Arcade <toolkit-name> Toolkit
+# Arcade your-toolkit-name Toolkit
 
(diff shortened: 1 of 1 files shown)


pages/toolkits/community-toolkit-template.mdx
```diff
@@ -1,3 +1,3 @@
-# Arcade <toolkit-name> Toolkit
```
sdreyer, line 1: Do you not want some sort of indication for placeholders?

EricGustin: I do

EricGustin: what would you suggest? The prod build is failing because of the previous placeholder <>

wdawson: Does escaping the carats work? `\<`

EricGustin: `&lt` and `&gt` would work, but I will just opt for UPPER-CASE as an indication for placeholder.

sdserranog (approved):
:shipit:

PR #175, opened 2025-02-27
Remove Govee, fix broken link, clarify pip installing for toolkits

EricGustin:


sdserranog:
Nice, thank you for removing the toolkit and adding the notes. Should we also remove close.io?

EricGustin:
@sdserranog i'll hold off on close.io for now until i have confirmation on whether the toolkit can be published or not

sdserranog (approved):
:shipit:

PR #179, opened 2025-03-03
Add signup event to PostHog with wrapped component.

wdawson:


pages/_app.tsx
```diff
@@ -17,7 +17,11 @@ function Docs({ Component, pageProps }) {
           process.env.NEXT_PUBLIC_POSTHOG_UI_HOST || "https://us.posthog.com",
         disable_session_recording: true,
         loaded: (posthog) => {
-          if (process.env.NODE_ENV === "development") posthog.debug();
+          if (
+            process.env.NODE_ENV === "development" ||
+            process.env.NEXT_PUBLIC_POSTHOG_DEBUG
```
wdawson, line 22: grabbed this idea from IdentityUI where you can toggle this to debug in any env

sdserranog:
@wdawson, could you provide some context for this? Is this a method to track where the users are coming from? If that’s the case, we are saving the UTMs in the events, and we can use something like this: https://posthog.com/tutorials/performance-marketing

wdawson:
@sdserranog this is [PostHog best practice](https://posthog.com/docs/getting-started/send-events#2-capture-custom-events) to send custom events in addition to the autocapture events for click events. This is because:

> 1. Allows us to send custom properties
> 2. Keeping events consistent over time

For this case ("Signup clicked" events), it's mostly the second one. We want to make sure the events are consistent as we change the text, location, etc. of the docs. We want to make sure we have a consistent funnel for how people are signing up. 

I wanted to preserve the UTM work you did, but this adds to it that tracking effort. Doesn't replace it.

src/components/NavBarButton.tsx
```diff
@@ -14,15 +12,12 @@ interface NavBarButtonProps {
...
-  href,
   text,
   variant = "default",
 }: NavBarButtonProps) => (
-  <Link href={href}>
-    <Button variant={variant}>
-      <span className="text-xs">{text}</span>
-    </Button>
-  </Link>
+  <Button variant={variant}>
```
sdserranog, line 18: I know this was like this before, but it is not valid HTML to have a button inside an anchor. We should either do something like [this](https://ui.shadcn.com/docs/components/button#link) or 
```jsx
<Button variant={variant} asChild>
    <Link href={href}>
      <span className="text-xs">{text}</span>
    </Link>
  </Button>
```

sdserranog:
> PostHog best practice

That sounds good to me! I wasn't sure if we wanted to monitor this as an event. The original purpose of UTMs was to determine where the users were coming from, and I believe the autocapture events were enough for this reason. But you are right. This is not replacing, just adding.

sdserranog (approved):
:shipit:

PR #183, opened 2025-03-06
Add Notion Auth Provider

EricGustin:

### DIFF
--- a/pages/home/auth-providers/index.mdx
+++ b/pages/home/auth-providers/index.mdx
@@ -62,6 +62,13 @@ Auth providers enable users to seamlessly and securely allow Arcade tools to acc
     category="Auth"
   />
   <ToolCard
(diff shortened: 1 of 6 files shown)


pages/home/auth-providers/index.mdx
```diff
@@ -61,6 +61,13 @@ Auth providers enable users to seamlessly and securely allow Arcade tools to acc
     link="/home/auth-providers/microsoft"
     category="Auth"
   />
+  <ToolCard
+    name="Notion"
+    image="notion"
```
wdawson, line 66: we need to add a notion image that's the same size as the other icons.

wdawson: lmk if you need help with that

wdawson (approved):

PR #187, opened 2025-03-17
Add Twitch and Reddit Auth Providers

EricGustin:


examples/code/integrations/twitch/custom_tool.py
```diff
@@ -0,0 +1,50 @@
...
+    duration: Annotated[
+        int,
+        "The duration of the poll in seconds.",
+    ],
+) -> Annotated[dict, "The poll that was created"]:
+    """Create a poll for a Twitch channel."""
+    url = "https://api.twitch.tv/helix/polls"
+    headers = {
+        "Authorization": f"Bearer {context.authorization.token}",
+        "Client-Id": "your_client_id",
```
EricGustin, line 37: I don't think I've ever seen the need for the client ID of the app to be in the header. Does the `ToolContext` contain this information?

pages/toolkits/entertainment/twitch.mdx
```diff
@@ -0,0 +1,118 @@
...
+
+## Configuring Twitch auth
+
+### Create a Twitch app
+
+- Twitch requires that you have two-factor authentication enabled for your account. Enable in your [account security seetings](https://www.twitch.tv/settings/security)
+- Create a Twitch Application in the [Twitch App Console](https://dev.twitch.tv/console/apps)
+- Set the OAuth Redirect URL to: `https://cloud.arcade.dev/api/v1/oauth/callback`
+- Select your Application category
+- Select the 'Confidential' Client Type
```
EricGustin, line 30: Should this be 'Confidential' client type or 'Public'?

Screenshot 2025-03-17 at 1 03 24 PM

nbarbettini: Confidential

examples/code/integrations/twitch/custom_tool.py
nbarbettini: Yes, this is very strange.
`ToolContext` does not contain the info, so it would have to be hardcoded for now. Filed: https://app.clickup.com/t/86b4a5bvj

nbarbettini (approved):

PR #189, opened 2025-03-18
Add Notion Toolkit Docs

EricGustin:


pages/toolkits/productivity/_meta.ts
```diff
@@ -1,6 +1,5 @@
 export default {
   google: "Google",
-  notion: "Notion",
```
EricGustin, line 3: no longer a folder, so not needed

nbarbettini (approved):

PR #219, opened 2025-04-01
Add instructions about the redirect url when configuring custom Oauth providers

byrro:
Our docs did not explain which redirect URL should be used when configuring a custom OAuth provider. Added that for cloud & self-hosted engine instructions.

### DIFF
--- a/pages/home/auth-providers/oauth2.mdx
+++ b/pages/home/auth-providers/oauth2.mdx
@@ -30,9 +30,15 @@ The only supported OAuth 2.0 flow is the authorization code grant flow (with or 
 
 How you configure the OAuth 2.0 provider depends on whether you use the Arcade Cloud Engine or a [self-hosted Engine](/home/install/overview).
 
-### Configuring OAuth 2.0 with the Arcade Dashboard
+### Arcade Cloud Engine
 
-1. Navigate to the OAuth section of the Arcade Dashboard and click **Add OAuth Provider**.
+#### Redirect URL
+
+The service you're configuring will require a redirect URL for the authorization flow. When using the Arcade Cloud, you should enter this URL: `https://cloud.arcade.dev/api/v1/oauth/callback`
+
+#### Add the OAuth Provider
+
+1. Navigate to the OAuth section of the [Arcade Dashboard](https://api.arcade.dev/dashboard/home) and click **Add OAuth Provider**.
 2. Select **OAuth 2.0** as the provider.
 3. Choose a unique **ID** for your provider (e.g. "my-oauth2-provider") with an optional **Description**.
 4. Enter your **Client ID** and **Client Secret** from your OAuth 2.0 provider.
@@ -40,7 +46,13 @@ How you configure the OAuth 2.0 provider depends on whether you use the Arcade C
 
 When you use tools that require OAuth 2.0 authorization using your Arcade account credentials, the Arcade Engine will automatically use this OAuth 2.0 provider.
 
-### Configuring OAuth 2.0 in self-hosted Arcade Engine configuration
+### Self-hosted Arcade Engine
+
+#### Redirect URL
+
+The service you're configuring will require a redirect URL for the authorization flow. When using a self-hosted Arcade Engine, you should enter this URL: `https://{your-engine-domain}/api/v1/oauth/callback`.
+
+> Replace `{your-engine-domain}` with the domain (or IP address) pointing to the server/cluster running the Arcade Engine.
 
 <Steps>
 


pages/home/auth-providers/oauth2.mdx
```diff
@@ -30,17 +30,29 @@ The only supported OAuth 2.0 flow is the authorization code grant flow (with or
 
 How you configure the OAuth 2.0 provider depends on whether you use the Arcade Cloud Engine or a [self-hosted Engine](/home/install/overview).
```
wdawson, line 31: You must always use the cloud address as the redirect URL. Can we revert this change but add something here like:

```suggestion
How you configure the OAuth 2.0 provider depends on whether you use the Arcade Cloud Engine or a [self-hosted Engine](/home/install/overview). If you use the Cloud Engine, you must configure your provider in the Dashboard. 

<Note>
When configuring your app in the OAuth 2.0 enabled service, you must use `https://cloud.arcade.dev/api/v1/oauth/callback` as the redirect URL (sometimes called callback URL).
</Note>
```

wdawson (changes requested):

pages/home/auth-providers/oauth2.mdx
```diff
@@ -28,19 +28,35 @@ The only supported OAuth 2.0 flow is the authorization code grant flow (with or
...
 When you use tools that require OAuth 2.0 authorization using your Arcade account credentials, the Arcade Engine will automatically use this OAuth 2.0 provider.
 
-### Configuring OAuth 2.0 in self-hosted Arcade Engine configuration
+### Self-hosted Arcade Engine
+
+#### Redirect URL
+
+The service you're configuring will require a redirect URL for the authorization flow. When using a self-hosted Arcade Engine, you should enter this URL: `https://{your-engine-domain}/api/v1/oauth/callback`.
+
+> Replace `{your-engine-domain}` with the domain (or IP address) pointing to the server/cluster running the Arcade Engine.
```
wdawson, line 59: Let's make sure to change this back

wdawson (changes requested):

pages/home/auth-providers/oauth2.mdx
byrro: updated

pages/home/auth-providers/oauth2.mdx
```diff
@@ -28,19 +28,23 @@ The only supported OAuth 2.0 flow is the authorization code grant flow (with or
...
-How you configure the OAuth 2.0 provider depends on whether you use the Arcade Cloud Engine or a [self-hosted Engine](/home/install/overview).
+How you configure the OAuth 2.0 provider depends on whether you use the Arcade Cloud Engine or a [self-hosted Engine](/home/install/overview). If you use the Cloud Engine, you must configure your provider in the Dashboard.
 
-### Configuring OAuth 2.0 with the Arcade Dashboard
+<Note>
+When configuring your app in the OAuth 2.0 enabled service, you must use `https://cloud.arcade.dev/api/v1/oauth/callback` as the redirect URL (sometimes called callback URL).
+</Note>
 
-1. Navigate to the OAuth section of the Arcade Dashboard and click **Add OAuth Provider**.
+### Arcade Cloud Engine
```
wdawson, line 37: We don't want to say Cloud since the dashboard is distributed with the engine and is available in self-hosted models.

```suggestion
### Using the Arcade Dashboard
```

pages/home/auth-providers/oauth2.mdx
```diff
@@ -28,19 +28,23 @@ The only supported OAuth 2.0 flow is the authorization code grant flow (with or
...
+1. Navigate to the OAuth section of the [Arcade Dashboard](https://api.arcade.dev/dashboard/home) and click **Add OAuth Provider**.
 2. Select **OAuth 2.0** as the provider.
 3. Choose a unique **ID** for your provider (e.g. "my-oauth2-provider") with an optional **Description**.
 4. Enter your **Client ID** and **Client Secret** from your OAuth 2.0 provider.
 5. Click **Save**.
 
 When you use tools that require OAuth 2.0 authorization using your Arcade account credentials, the Arcade Engine will automatically use this OAuth 2.0 provider.
 
-### Configuring OAuth 2.0 in self-hosted Arcade Engine configuration
+### Self-hosted Arcade Engine
```
wdawson, line 47: We should revert this since we want to explain how to use the engine.yaml file for self-hosted engines. You can still use the dashboard for self-hosted engines. Maybe we could do something like:

```suggestion
### Using the `engine.yaml` configuration file

<Note>
This method is only available when you are [self-hosting the engine](/home/install/overview)
</Note>
```

pages/home/auth-providers/oauth2.mdx
```diff
@@ -28,19 +28,27 @@ The only supported OAuth 2.0 flow is the authorization code grant flow (with or
...
 
-### Configuring OAuth 2.0 with the Arcade Dashboard
+<Note>
+When configuring your app in the OAuth 2.0 enabled service, you must use `https://cloud.arcade.dev/api/v1/oauth/callback` as the redirect URL (sometimes called callback URL).
+</Note>
 
-1. Navigate to the OAuth section of the Arcade Dashboard and click **Add OAuth Provider**.
+### Using the Arcade Dashboard
+
+1. Navigate to the OAuth section of the [Arcade Dashboard](https://api.arcade.dev/dashboard/home) and click **Add OAuth Provider**.
```
wdawson, line 39: we need to have a different url for self-hosted or just have a link for the cloud.

byrro: this is under `### Arcade Cloud Engine`, so I guess we should only have the cloud URL here?

byrro: What would you suggest, @wdawson? I'm confused about which heading and what to put in the instructions

wdawson: I might drop the link entirely (like it was before). Or something like
```suggestion
1. Navigate to the OAuth section of the Arcade Dashboard ([here](https://api.arcade.dev/dashboard/home) if using Arcade cloud) and click **Add OAuth Provider**.
```

Over time, maybe we need a docs page about how to access the dashboard during a self-hosted installation

byrro:
@wdawson I lost track of this. Just committed your latest suggestion. Also added explanations on how to access self-hosted dashboard and where to locate the engine.yaml file. Let me know your thoughts

wdawson (approved):

PR #222, opened 2025-04-01
Remove Last Updated Timestamp 

EricGustin:
The last updated timestamp on docs is misleading. "Last updated on March 21, 2025" at the bottom of every page seems to apply to the whole site, not the particular page you're looking at.

Nextra needs the full git commit history in order to provide accurate 'Last updated on'. When we deploy, we're only providing a shallow clone, so only the most recent commit date is available to Nextra. This is why the issue only exists in production and not locally.

ClickUp ticket: https://app.clickup.com/t/86b4d4g2a

### DIFF
--- a/pages/home/supported-models/_meta.ts
+++ b/pages/home/supported-models/_meta.ts
@@ -1,3 +1,7 @@
 export default {
   index: "Overview",
+  anthropic: "Anthropic",
+  groq: "Groq",
+  ollama: "Ollama",
+  openai: "OpenAI",
 };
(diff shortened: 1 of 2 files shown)


pages/home/supported-models/_meta.ts
```diff
@@ -1,3 +1,7 @@
 export default {
   index: "Overview",
+  anthropic: "Anthropic",
+  groq: "Groq",
+  ollama: "Ollama",
+  openai: "OpenAI",
```
EricGustin, line 6: sneaking this in so that OpenAI is rendered as "OpenAI" instead of "Openai" on the sidebar

wdawson (approved):

PR #224, opened 2025-04-02
Third round of docs review covering Jakub's suggestions

byrro:
This PR covers the remaining issues pointed by Jakub.

### DIFF
--- a/pages/home/auth/call-third-party-apis-directly.mdx
+++ b/pages/home/auth/call-third-party-apis-directly.mdx
@@ -23,8 +23,8 @@ This can be useful when you need to manage authorization flows in your applicati
 - Sign up for an <SignupLink utmParams={{
             utm_source: "docs",
             utm_medium: "page",
-            utm_campaign: "custom-tools",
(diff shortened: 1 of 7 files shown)


pages/home/auth/call-third-party-apis-directly.mdx
```diff
@@ -19,8 +20,14 @@ This can be useful when you need to manage authorization flows in your applicati
 
 ### Prerequisites
 
-- [Set up Arcade](/home/quickstart)
-- Install the Arcade Client and the required libraries:
+- Sign up for an <SignupLink utmParams={{
+            utm_source: "docs",
+            utm_medium: "page",
+            utm_campaign: "custom-tools",
```
wdawson, line 26: probably want a different string here? seems like this isn't the custom tools page

wdawson (approved):
One small tweak to the utm params, but looks good to me otherwise.

pages/home/auth/call-third-party-apis-directly.mdx
byrro: fixed

PR #228, opened 2025-04-17
remove unused publish.yaml

evantahler:
I'm pretty sure we don't use this (there's no master branch), and we deploy to cloudflare anyway


EricGustin (approved):
nice catch

PR #232, opened 2025-04-20
Document the Salesforce toolkit and auth provider configuration

byrro:


pages/home/auth-providers/salesforce.mdx
```diff
@@ -0,0 +1,367 @@
...
+
+This auth provider is used by:
+
+- The [Arcade Salesforce toolkit](/toolkits/sales/salesforce), which provides pre-built tools for interacting with Salesforce services
+- Your [app code](#calling-salesforce-apis-directly) that needs to call Salesforce APIs
+- Or, your [custom tools](#create-your-own-salesforce-tools) that need to call Salesforce APIs
+
+## Create a Salesforce app
+
+Follow Salesforce's guide to [creating your Connected App](https://help.salesforce.com/s/articleView?id=xcloud.connected_app_create.htm&type=5).
```
sdserranog, line 24: Not important because I could edit after creating the app, but I entered the link without reading the next steps. Later, I realized I needed to check the enabled OAuth settings and set the callback URL, etc. Perhaps we can add something here to prevent that, but I'm not sure what.

pages/home/auth-providers/salesforce.mdx
```diff
@@ -0,0 +1,367 @@
...
+
+## Get your Salesforce Org Subdomain
+
+Follow the steps below to find your Salesforce Org Subdomain:
+
+1. Log into your Salesforce account.
+1. Click on the **Setup** gear icon in the top right corner, then click on **Setup**.
+1. In the **Quick Find** box in the top left corner, type `"my domain"`.
+1. In the search results, under **Company Settings**, click on **My Domain**.
+1. Under **My Domain Details**, check the **Current My Domain URL** field.
```
sdserranog, line 44: Not sure if this screen is different in other accounts, but shouldn't it be easier to simply say, use `My domain name` instead of all this description?

Screenshot 2025-04-23 at 11 53 17 PM

examples/code/integrations/salesforce/config_provider.engine.yaml
```diff
@@ -0,0 +1,30 @@
...
+  providers:
+    - id: salesforce
+      description: "Custom Salesforce provider"
+      enabled: true
+      type: oauth2
+      client_id: ${env:SALESFORCE_CLIENT_ID}
+      client_secret: ${env:SALESFORCE_CLIENT_SECRET}
+      oauth2:
+        authorize_request:
+          endpoint: 'https://{salesforce-org-subdomain}.my.salesforce.com/services/oauth2/authorize'
```
sdreyer, line 11: Feel like this is going to mess people up. Since we use templates in the rest of this like '{{client_id}}' i assumed {salesforce-org-subdomain} was a template and it gave me an error I had to think about for a sec. 

Feel like someone unfamiliar is going to overlook that

byrro: that makes sense.. perhaps we can use brackets? or asterisks? do you have a good idea?

pages/home/auth-providers/salesforce.mdx
byrro: yea, it's not always the same.. if you have a developer account, for instance, it'll be different. not sure if there are other cases

pages/home/auth-providers/salesforce.mdx
byrro: rephrased it to mention our instructions before introducing the salesforce link

examples/code/integrations/salesforce/config_provider.engine.yaml
byrro: changed to simply `https://salesforce-org-subdomain.my.salesforce.com/services/oauth2/authorize`

in the docs, every time the domain is mentioned, I remind the reader to replace `salesforce-org-subdomain` with the actual one

sdserranog (approved):
LGTM!

PR #237, opened 2025-04-24
Remove tool instalation header from docs

evantahler:
The persona we want to focus on is the agent developer, using the cloud product.  This means we should remove the 'install' step from the docs.  Instead, I added a footer with an 'escape hatch' that takes you either to the cloud docs or the self-hosting docs.  Folks /shouldn't/ be installing the pip packages 

<table>
  <tr>
    <td>
Screenshot 2025-04-24 at 3 27 45 PM
 </td>
    <td>
Screenshot 2025-04-24 at 3 27 22 PM
</td>
  </tr>
</table>


EricGustin (approved):
Agent Makers Unite

PR #240, opened 2025-04-29
Streamlined quickstart example

wdawson:
Devs can now copy/paste in one go instead of needing to parse three disconnected snippets.


evantahler (approved):
Copy/pasteable customer success!

PR #241, opened 2025-04-29
Use new logo

evantahler:


sdserranog (approved):
This is great, not blocking. But we could also change the `favicon`, and adjust the padding for the logo, too. 
Screenshot 2025-04-29 at 2 34 17 PM

src/components/Logo.tsx
```diff
@@ -1,7 +1,7 @@
 export function Logo() {
   return (
     <>
-      <div className="mr-16 hidden w-40 items-center md:block">
+      <div className="mr-2 hidden w-40 items-center md:block">
```
evantahler, line 4: @sdserranog I think a /little/ margin looks good?

PR #244, opened 2025-05-01
Add MCP Docs

EricGustin:
Local:
  - Claude Desktop
  - Python Client
  - TypeScript Client
 
Remote
  - Claude Desktop
  - TypeScript Client
  
-----
  
  An example of using the demo mcp worker is still needed.


examples/code/home/mcp/streamable-http/typescript-client.ts.fake
EricGustin, line 1: need to do `.ts.fake` file extension, else the linter complains

examples/code/home/mcp/streamable-http/typescript-client.ts.fake
```diff
@@ -0,0 +1,129 @@
...
+        // Re-throw other errors
+        throw error;
+      }
+    }
+  }
+
+  await callTool();
+} catch (error) {
+  console.log("Error calling tool:", error);
+}
```
EricGustin, line 125: TODO: simplify?

examples/code/home/mcp/streamable-http/typescript-client.ts.fake
```diff
@@ -0,0 +1,129 @@
...
+
+// Exchange API key for user-scoped token
+const tokenExchangeResponse = await fetch(`${arcadeURL}/tokens/exchange`, {
+  method: "POST",
+  headers: {
+    Authorization: `Bearer ${arcadeApiKey}`,
+    "Content-Type": "application/json",
+  },
+  body: JSON.stringify({ user_id: userId }),
+});
```
EricGustin, line 24: will we be exposing this endpoint in our arcadejs SDK?

pages/home/mcp/local/code-client.mdx
```diff
@@ -0,0 +1,200 @@
...
+```bash
+pip install mcp
+pip install arcade-ai
+pip install arcade-google
+```
+</Tabs.Tab>
+
+<Tabs.Tab>
+```bash
+npm install @modelcontextprotocol/sdk
```
EricGustin, line 32: if they wanted to write a typescript client for stdio, then they would still need to pip install arcade-ai which feels odd

nbarbettini:
I made a few changes:
- Removed the local & remote folders and hid some examples (for now) so the reader just has to decide between "I want to use Arcade with Claude (locally)" and "I want to use Arcade from my code (remotely)". These 2 scenarios map onto what is best supported today, as a starting point. I expect it will change quickly as the MCP ecosystem evolves.
- Also hid the remote TypeScript example -- pending prod deploy

The net result is that `MCP > Arcade with Claude Desktop` is the only page currently visible. It feels like that is worth deploying all by itself, and then I will open a new PR immediately after that support is on prod, to re-enable the remote Typescript example.

Thank you for busting out these docs @EricGustin!

nbarbettini (approved):

PR #251, opened 2025-05-07
2025 docs revamp integration branch

evantahler:
TODO:

- [x] Re-order section based on expected flow and value prop
- [x] Remove the LLM/Model APIs entirely
- [x] Quickstart is now about calling tools directly
- [x] Overview page is shorter and makes our valueprop clear
- [x] local self hosting talks about how our cloud server is still used
- [x] Section: Arcade CLI
- [x] Section Guides
- [x] Make tool-building section simpler and remove notion of worker
- [ ] ~~Replace self-hosting with "contact us" section~~


Spartee:
We don't have a llamaindex integration yet.

examples/code/home/use-tools/call-tools-directly/math_sqrt_directly.py
```diff
@@ -1,13 +1,41 @@
...
+# Now, let's use a tool that requires authentication to star a GitHub repository
+
+auth_response = client.tools.authorize(
+  tool_name="GitHub.SetStarred",
+  user_id=user_id,
+)
+
+if auth_response.status != "completed":
+    print(f"Click this link to authorize: {auth_response.url}")
+    print("re-run the script after authorizing")
```
torresmateo, line 28: Wondering if it would be better to introduce calling `client.auth.wait_for_completion` here? It feels like a nice way to have a quick "nice one!" early on.

evantahler: Great idea! a9f1dd57ea958880e238c382fcf37a1b226689eb

wdawson (approved):

PR #253, opened 2025-05-07
Toolkit development needs a virtualenv (CU-86b4xz6dc)

evantahler:


pages/home/build-tools/create-a-toolkit.mdx
```diff
@@ -15,7 +15,10 @@ Before starting, make sure you have:
 
 - Python 3.10 or higher (verify with `python --version`)
 - An [Arcade account](https://api.arcade.dev/signup?utm_source=docs&utm_medium=page&utm_campaign=custom-tools)
-- Arcade CLI installed: `pip install arcade-ai`
+- An active virtual environment for your shell(s).
+    - We recommend using [uv](https://docs.astral.sh/uv/pip/environments/) or [miniconda](https://www.anaconda.com/docs/getting-started/miniconda/main).
+    - Toolkits each have a poetry project with their own virtual environment, but toolkits are loaded into the top-level arcade environment based on the active  "parent" virtual environment.  **The local development commands will not work without an active virtual environment.**
+- Arcade CLI installed within your python virtual environment: `pip install arcade-ai`
```
wdawson, line 21: Should this have separate instructions depending on the kind of venv? E.g. `uv tool install arcade-ai` or `uv pip install arcade-ai`?

It's unclear which we should recommend to me. `uv`'s pip interface appears to be for "legacy workflows"

https://docs.astral.sh/uv/getting-started/features/#tools

evantahler: I... don't know?  What do you think we should do?

wdawson: I think this change as is, is probably better than what we have. We can iterate on it like you had in the CLI docs I think for the revamp branch.

PR #256, opened 2025-05-08
Add CLI docs page

evantahler:
Adds a page about the CLI in general.  @wdawson I added some detail here about how to setup a python virtual environment a few different ways here. 

I was kind of at a loss for what to put for content here, so I just dumped the help text?

Screenshot 2025-05-07 at 5 34 35 PM

Of note, I moved this into the "core concepts" section - I don't think it needs to be a top-level section any more


sdserranog:
I've always been curious about this: is it common to install a Python package that is a CLI in a specific Python virtual environment? Almost all the CLIs I've used (aws, vercel, docker etc) are global, and you install them on Mac, for example, with either `brew` or `npm install -g` (which is global).

evantahler:
@sdserranog I'm with you... it doesn't feel right.  That's probably why one the many reasons that other that AI/data, all developer tools are moving to other languages ;p 

@nbarbettini / @wdawson can you check me on this?  I know that our local development commands (`arcade show --local`, etc) don't do much without a virtualenv loaded up with tools...

wdawson:
I would like our CLI to eventually be a `brew install` for macOS and other equivalent installers for other operating systems. Maybe it's written in python, i don't really care about that too much.

It does feel weird to require python to use Arcade. But, for now, with the primary capability of the CLI being to develop tools which can only occur in python right now, I don't think it's a huge issue ATM.

Good opportunity for painted doors I think for "installation methods" (brew, apt, ...) and also building tool languages (python, js/ts, ...)

pages/home/arcade-cli.mdx
```diff
@@ -0,0 +1,78 @@
...
+
+Now that your python virtual environment is activated, you can install the Arcade CLI with the following command:
+```bash
+pip install arcade-ai
+```
+
+## Usage
+
+```bash
+$ arcade --help
```
wdawson, line 49: probably fine for now. I would like to graduate to something like: https://ngrok.com/docs/agent/cli/

OR we could have `--help` be the cannonical place. Mostly worried about things drifting.

pages/home/arcade-cli.mdx
```diff
@@ -0,0 +1,78 @@
...
+
+import { Tabs } from "nextra/components";
+
+# The Arcade CLI
+
+The Arcade CLI is a command-line tool that allows you to manage your Arcade deployments, generate, test, and manage your toolkits, and more.
+
+## Installation
+
+Like all python packages, the Arcade CLI needs to be installed within the python virtual environment you are using for your Arcade deployment.
```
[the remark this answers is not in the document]
evantahler: fixed in d7e9fd9a096b9d280a7feda0d2340cfcf9690458

pages/home/arcade-cli.mdx
evantahler: For now, I'll list all the command's help for SEO I guess

pages/home/arcade-cli.mdx
```diff
@@ -0,0 +1,334 @@
...
+
+conda create --name arcade
+conda activate arcade
+```
+</Tabs.Tab>
+<Tabs.Tab>
+```bash
+# install uv: https://docs.astral.sh/uv/getting-started/installation/
+
+uv venv
```
wdawson, line 35: I think we need `--seed` here to also include `pip`
```suggestion
uv venv --seed
```

pages/home/arcade-cli.mdx
```diff
@@ -0,0 +1,334 @@
...
+uv source .venv/bin/activate
```
wdawson, line 36:
```suggestion
source .venv/bin/activate
```

pages/home/arcade-cli.mdx
```diff
@@ -0,0 +1,334 @@
...
+```
+
+You can learn more about any of the commands by running `arcade <command> --help`, e.g. `arcade new --help`.
+
+## `arcade login`
+
+```bash
+ Usage: arcade login [OPTIONS]
+
+ Log in to Arcade Cloud
```
sdserranog, line 83: It may be worth mentioning that it also generates your Arcade API key, and this is a way to obtain it.

sdserranog: I know this is our current description, but maybe it's worth mentioning here

wdawson: I'd rather not send people there to find it. It could change later

wdawson (approved):

pages/home/arcade-cli.mdx
wdawson: Also there's this: https://docs.arcade.dev/home/api-keys

PR #258, opened 2025-05-08
GHA for rebuilding LLMs.txt

evantahler:
Because of both the need for the site to be public, and the cost of running this, We can use a github action that rebuilds the `llms.txt` files once a week, from the production site.  Any changes will be committed back to main and then re-deployed.

This action can also be run manually.


evantahler:
FIRECRAWL_API_KEY set as repository secret

EricGustin (approved):

PR #261, opened 2025-05-08
Move some auth stuff around

wdawson:
Looks like there also might have been a bad force push so fixing some of it for @evantahler


pages/home/auth-providers/index.mdx
```diff
@@ -9,6 +9,19 @@ import { ToolCard } from "@/components/ToolCard";
 
 Auth providers enable users to seamlessly and securely allow Arcade tools to access their data.
 
+Arcade has several auth providers available in the Arcade cloud so you don't have to configure your own.
+However, using Arcade's auth providers means that your users will see the Arcade brand (name and logo) on
+the auth screen and your authentications will share any rate limits from those providers with other Arcade
+customers.
```
nbarbettini, line 15:
```suggestion
Arcade has many auth providers available in Arcade Cloud so you don't have to configure your own.
However, using Arcade's default auth providers means that your users will see the Arcade brand (name and logo) on
the auth screen and your authentications will share any rate limits from those providers with other Arcade
customers.
```

some nits, if you agree

nbarbettini: The only change^ I feel strongly about is `Arcade Cloud` as a proper name.

pages/home/_meta.tsx
```diff
@@ -72,6 +72,9 @@ export default {
   "serve-tools": {
     title: "Serve tools",
   },
+  "auth-providers": {
+    title: "Customizing Auth",
+  },
```
nbarbettini, line 77: It doesn't feel like this fits in Building Tools. Maybe under Using Arcade? It's about using our platform, not building tools specifically.

evantahler (approved):
There are 2 reasons to get to these pages - either you want to self-host the engine, or you want to change from our default oAuth apps to your own.  Because of the second case,  I think this is the right place?

pages/home/_meta.tsx
wdawson: @nbarbettini let me know what you think of @evantahler 's comment below. I'm pretty sure they don't belong in Self-Hosting. Maybe I should add that as a reason why you'd need your own apps though. 

I could see up in Using Arcade, but that feels too high up in the Nav for people to worry about that concept. I thought about in Core Concepts, but didn't like an ever-growing list being there. Integrations would be nice, but it's not geared for more than toolkits for now.

Trying to find the goldilocks zone.

evantahler (approved):
Let's do it!

PR #266, opened 2025-05-10
Remove OpenAI LLM tool calling examples from toolkits in docs

evantahler:
re: https://arcade-ai.slack.com/archives/C07EXGYH91B/p1746838868812149

This is part of our work to remove the LLM/Model APIs.  I think I'd like to make an argument for removing these types of examples - I don't think they provide that much value after all. The direct-call examples are better. Consider this typical example:

```js
import OpenAI from 'openai';

const USER_ID = "you@example.com";
// TODO: Replace with an identifier for a real post
const PROMPT = "Get the top level comments of the post https://www.reddit.com/r/TestSubreddit/comments/1abcdefg/"
const TOOL_NAME = "Reddit.GetTopLevelCommentsOfPost";

const client = new OpenAI({
  baseURL: 'https://api.arcade.dev/',
  apiKey: process.env.ARCADE_API_KEY
});

const response = await client.chat.completions.create({
  messages: [
    { role: 'user', content: PROMPT }
  ],
  model: 'gpt-4o-mini',
  user: USER_ID,
  tools: [TOOL_NAME],
  tool_choice: 'generate'
});

console.log(response.choices[0].message.content);
```

The prompt is new and interesting, but the rest isn't:
* the tool name you know because you are in that section
* we don't want you to make openAI clients any more
* you don't see an example response, so that's not interesting either


EricGustin (approved):
￼

PR #270, opened 2025-05-12
create-a-toolkit preamble

evantahler:


pages/home/build-tools/create-a-toolkit.mdx
```diff
@@ -9,6 +9,12 @@ import { Steps, Tabs } from "nextra/components";
 
 This guide walks you through the complete process of creating a custom toolkit for Arcade - from initial setup to publishing on PyPI. You'll build a simple arithmetic toolkit with operations like addition, subtraction, and multiplication.
 
+## Workflow Overview
```
wdawson, line 12: Can we combine this with the previous section? I don't know that we need a whole new header? It also seems like we have several introductory sentences in a row as a result instead of a cohesive introduction.

wdawson (approved):

PR #274, opened 2025-05-13
Add Clients page

evantahler:
Also, I made a go Client too.  Thanks, Stainless! https://github.com/ArcadeAI/arcade-go

### DIFF
--- a/pages/home/arcade-sdks.mdx
+++ b/pages/home/arcade-sdks.mdx
@@ -23,9 +23,4 @@ Learn more about the [JavaScript / TypeScript SDK](https://github.com/ArcadeAI/a
 
 Install with: `go get -u 'github.com/stainless-sdks/arcade-engine-go'`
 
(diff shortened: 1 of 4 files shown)


pages/home/arcade-sdks.mdx
```diff
@@ -0,0 +1,26 @@
...
+
+## JavaScript / TypeScript SDK
+
+Install with: `npm install @arcadeai/arcadejs`
+
+Learn more about the [JavaScript / TypeScript SDK](https://github.com/ArcadeAI/arcade-js).
+
+## Go SDK
+
+Install with: `go get -u 'github.com/stainless-sdks/arcade-engine-go'`
```
wdawson, line 24: Why is this `arcade-engine`? Is it only for the engine APIs?

wdawson: Should we fix stainless to be consistent with the other SDKs?

evantahler: yeah, and it's not pointing to our public github either... digging

PR #284, opened 2025-05-16
Fix some small linting things

wdawson:
Don't think we're running linting correctly, but I'm not sure how to fix the root cause. These changes now mean that `pnpm lint` passes with no issues


sdserranog (approved):
It's probably an issue with `husky`

PR #294, opened 2025-05-22
fixed broken internal links in the 'Hosting options' section

torresmateo:
Some of the internal links on the "Hosting options" section are currently broken, this PR restores these links.


evantahler (approved):
Thank you!

PR #296, opened 2025-05-22
Clarify docs for auth provider choice

wdawson:
We don't want to suggest people name (set the `id` of) their providers to be the same as the `provider_id`. That yields more of a special case behavior than what we've intended. 

Also, I don't want to sprinkle this too much since I think it could cause more confusion for the 80% (we think) of people that won't have multiple providers of the same type. 

Proposing this as a happy medium for now.


torresmateo (approved):
LGTM, maybe the FAQ entry should point to `/home/auth-providers#using-multiple-providers-of-the-same-type` or be removed. This explanation is more complete!

wdawson:
Sweet! Removed from the FAQ. Hopefully won't be a frequent question!

PR #311, opened 2025-06-03
added Google ADK section

torresmateo:
This PR:
- orders the framework list alphabetically (happy to change to a custom order if preferred)
- adds docs to the new `google-adk-arcade` library


evantahler (changes requested):
Other than the sidebar order, 👍

PR #313, opened 2025-06-06
Fix so many broken links

EricGustin:
So yeah, this is bad.

### Additional notes:
The breadcrumbs for toolkit pages are broken. Not sure how to fix this.


evantahler (approved):
Let's get these merged now.  I'll add a link checker as a test suite (`next-validate-link`) in another PR

PR #316, opened 2025-06-09
Tool auth docs

sdreyer:

### DIFF
--- a/pages/home/auth/tool-auth-status.mdx
+++ b/pages/home/auth/tool-auth-status.mdx
@@ -153,8 +153,7 @@ const userId = "you@example.com";
 const toolName = "Google.ListEmails";
 
 // Get specific tool details
-const tool = await client.tools.get({
-    tool_name: toolName,
+const tool = await client.tools.get(toolName, {
     user_id: userId
 });
 


sdserranog:
This is amazing! Can we upgrade the [reference](https://reference.arcade.dev/api-reference#tag/tools/GET/v1/tools) and add it to our docs? This way, people can see what the endpoint returns directly

sdreyer:
> This is amazing! Can we upgrade the [reference](https://reference.arcade.dev/api-reference#tag/tools/GET/v1/tools) and add it to our docs? This way, people can see what the endpoint returns directly

@evantahler how are these updated again? Is it automatic?

pages/home/auth/tool-auth-status.mdx
```diff
@@ -0,0 +1,208 @@
...
+<Tabs.Tab>
+```js
+const userId = "you@example.com";
+const toolName = "Google.ListEmails";
+
+// Get specific tool details
+const tool = await client.tools.get({
+    tool_name: toolName,
+    user_id: userId
+});
```
sdserranog, line 159: This is how we specify a tool in the JS client:
```suggestion
const tool = await client.tools.get(toolName, {
    user_id: userId
});
```

evantahler:
> @evantahler how are these updated again? Is it automatic?

Should be!  Whatever ends up on the swagger file in the engine gets deployed

pages/home/auth/tool-auth-status.mdx
sdreyer: ty, did I miss anything else in the JS docs?

sdserranog (approved):

PR #317, opened 2025-06-10
Fix more broken links

evantahler:
Solves a bunch of issues from an ahrefs audit

Screenshot 2025-06-09 at 5 48 48 PM

Of note, the way our breadcrumbs work, the mid-category link goes to the first toolkit in the directory.  Alphabetizing the links made us LUCKY, in that no item with sub-folders was first.  We will need to fix this, but later...

Screenshot 2025-06-09 at 5 47 08 PM


EricGustin:
Can we hide the inde.mdx from the sidebar. This shouldn't break the functionality you're adding, but makes the sidebar prettier.

Example _meta.ts:
```ts
export default {
  index: {
    display: "hidden",
  },
  calendar: "Calendar",
  contacts: "Contacts",
  docs: "Docs",
  drive: "Drive",
  gmail: "Gmail",
  sheets: "Sheets",
  reference: "Reference",
};
```

image

EricGustin (approved):
noice

PR #329, opened 2025-06-18
surfaced toolkit names

torresmateo:
This adds the toolkit name to each tool heading so it's easy to avoid naming issues


byrro (approved):
LGTM

PR #340, opened 2025-07-10
Update the Slack toolkit docs for new refactored version

byrro:


EricGustin (approved):
love the deprecation messages

PR #358, opened 2025-07-21
Clarify provider ID necessary to make a custom Atlassian provider work with Jira & Confluence toolkits

byrro:


byrro:
Closing it, as since it's selected as an "included provider", it'll always have the correct provider id, regardless of the ID input by the user.

PR #365, opened 2025-07-26
Glossary val changes

vfanelle:


evantahler:
@vfanelle do you need help updating this PR?

PR #384, opened 2025-07-31
[Jira ArcToolkit][ENG-1537] Adding documentation for new tools GetBoards and ListSprintsForBoards

jottakka:
[Clickup: ENG-1537](https://app.clickup.com/t/9014390315/ENG-1537)

Adding documentation based on what was generated using the  `arcade docs` tool with some tweak
Adding docs for those tools:
- Jira.GetBoards
- Jira.ListSprintsForBoards

PS: Thanks to @torresmateo for the lesson on how to generate those! This guy is great!
PSS: The tool is great, maybe having the option to list the tools that we want to add will spare some tokens, but nothing that taking a good look in the git diff would not solve.


torresmateo:
This looks good, the only nit is that the github link is broken, but I think this relies on #370

torresmateo (approved):

PR #405, opened 2025-08-10
Docs: GoogleSlides, GoogleSheets, GoogleDocs

EricGustin:
## GoogleSheets (merged & on staging):
* SearchSpreadsheets - Lets the agent discover which spreadsheet files exist
* UpdateCells - Let's the agent update multiple ranges of cells in a single tool call.
* GetSpreadsheetMetadata - Returns, for a given spreadsheet ID, the list of every tab (sheet) inside it (sheet IDs, names, positions, row-/column counts).
* AddNoteToCell - Add a note to a specific cell in a spreadsheet. Unfortunately, adding comments to a specific cell is not supported by the API. Using notes was my compromise. They're very similar.

## GoogleDocs (merged & on staging):
* CommentOnDocument - Comment on a Google Doc. This is not anchored to a specific page/sentence, but is a comment on the document itself. Commenting on a single page/sentence/word is not supported by the Google API.
* ListDocumentComments - Lists all of the comments & replies that are in the document.

## GoogleSlides (awaiting review/approval):
* CreatePresentation - Creates a new presentation & fills out the first slide with the title and subtitle
* CreateSlide - Creates a new slide at the end of the presentation. Supports the "Title and Body" template.
* GetPresentationAsMarkdown - Gets the whole presentation's text content as markdown. Supports text styling like bold, italics, underlines, hypertext links, and strikethroughs. Also supports bulleted lists & layered bulleted lists.
* SearchPresentations - Search for presentations in your google drive
* CommentOnPresentation - Comment on a presentation. This is not anchored to a specific slide, but is a comment on the presentation itself. Commenting on a single slide is not supported by the Google API.
* ListPresentationComments - Lists all of the commens & replies that are in the presentation.


evantahler (changes requested):
Google slides should be added to https://docs.arcade.dev/toolkits too!

EricGustin:
> Google slides should be added to https://docs.arcade.dev/toolkits too!

good catch! updated

PR #408, opened 2025-08-15
fixed broken links in integration cards

torresmateo:
Some toolkits on the index had broken links.


nbarbettini (approved):
thanks @torresmateo!

PR #410, opened 2025-08-15
new overview image to prep for MCP local world

evantahler:
Before:
before

After:
Screenshot 2025-08-18 at 9 12 31 AM


arcadeben (commented):
As we add things like tool selection, where does that fit on this diagram?  I'm assuming it's a horizontal bar similar to auth

evantahler:
> As we add things like tool selection, where does that fit on this diagram?  I'm assuming it's a horizontal bar similar to auth 

Could be! Or we collapse it to have all of our platform features as callouts

PR #447, opened 2025-09-20
chore: Update OAuth callback instructions

nbarbettini:
Updates all of our OAuth app instructions to reflect the fact that each app has a unique redirect URL.


evantahler (approved):
I'm not sure if we should muddy these pages for self-hosters, but that's another topic for another day

nbarbettini:
> another topic for another day

Agreed: https://app.clickup.com/t/9014390315/ENG-1702

PR #464, opened 2025-10-02
Glossary terms popover

evantahler:
We've got a glossary that defines our terms-of-art.  An idea from @vfanelle was to explain them to people throughout the docs.  

Screenshot 2025-10-01 at 5 42 37 PM

This uses a remark plugin to inject popovers into the build process for the markdown pages, but that didn't work with turbopack, so I removed turbopack  

The experience of adding adding new terms remains as it did before - just add them to the glossary markdown page.

Of note, this only works for the english language so far.


sdserranog (approved):
This is really nice! It looks a lot better now

PR #468, opened 2025-10-02
Get toolkits from the design system.

sdserranog:
This is the initial effort to eliminate the hardcoded list of toolkits from our docs and establish a single source of truth.

- Grab the icons and toolkits from the design system
- Improve filters
- Add BYOC and isPRO badge


evantahler:
I have a request - now that the design system is in place, can the detail pages for the toolkits include their icon and badges?

sdserranog:
> I have a request - now that the design system is in place, can the detail pages for the toolkits include their icon and badges?


I will add it to my next bunch of tasks for docs!

PR #476, opened 2025-10-06
Use relative links and enforce it

evantahler:
Using http:// links to ourself is slow and breaks hydration


sdserranog (approved):
Nice 
:shipit:

PR #478, opened 2025-10-07
Use new logo.

sdserranog:
Screenshot 2025-10-06 at 10 24 15 PM

New one:
Screenshot 2025-10-07 at 6 30 08 PM


evantahler:
I don't know how to describe this properly... but it looks odd that "docs" isn't in the arcade font to me, especially in the header.  I won't block the PR, but is there some way to use the logo font here?

evantahler (approved):

PR #527, opened 2025-10-22
Implement `make toolkit-docs` auto-docs builder command

byrro:
Changed it to run as `make toolkit-docs` instead of a Python CLI.

The inputs are now required in an interactive way through the terminal, which is a better UX.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ run: ## Run the docs site locally
 	@pnpm dev
 
 toolkit-docs: ## Generate documentation for an MCP Server toolkit
-	@uv run python -m make_toolkit_docs
+	@cd make_toolkit_docs && uv sync && uv run python __main__.py
 
 ruin:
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -18,6 +18,9 @@ test: ## Run the tests
 run: ## Run the docs site locally
 	@pnpm dev
 
+docs: ## Generate documentation for a server (interactive)
```
EricGustin: I think make toolkit-docs would be less confusing (given that the repo itself is named docs)

Makefile
```diff
@@ -18,6 +18,9 @@ test: ## Run the tests
...
+	@cd make_docs && uv run python -c "from __init__ import run_command; run_command()"
```
EricGustin: a `__main__.mpy` file would simplify it to `uv run python -m make_docs`

PR #529, opened 2025-10-22
compare server types

torresmateo:
This adds a conceptual page to help people visualize the different types of MCP servers they can use with Arcade, as well as the different feature suppport depending on what they choose.

### DIFF
--- a/app/en/home/compare-server-types/page.mdx
+++ b/app/en/home/compare-server-types/page.mdx
@@ -2,28 +2,16 @@
 title: "Compare Server Types"
 description: "Compare the different types of MCP servers"
 ---
+
 # Compare MCP Server Types
 
 Depending on the transport you use and where you want to run your MCP server, Arcade offers different functionalities and features. Below is a comparison of the different server types and their capabilities.
 
-Server types:
-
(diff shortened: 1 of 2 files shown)


app/en/home/compare-server-types/page.mdx
```diff
@@ -0,0 +1,28 @@
...
+
+Server types:
+
+| Transport | Deployment                      | Description                                                                                                                                                                                               |
+| --------- | ------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| stdio     | local                           | MCP Server running as a subprocess of the client. Local only.                                                                                                                                             |
+| http      | local (unprotected)             | MCP server running on a local network, accessible from the local machine, without server-level auth.                                                                                                      |
```
EricGustin: we could add these terms (protected vs unprotected) to our glossary so that they can hover on it on this page and see what they mean. Could be useful to link out to the official MCP specification/docs

torresmateo: Based on the slack conversation, I moved the definitions to the glossary, and removed this table

PR #531, opened 2025-10-23
collapse sidebar on home page to keep focus on content, other pages have it open

evantahler:
I'm emulating clicks to trigger the animations.

### DIFF
--- a/app/_components/sidebar-controller.tsx
+++ b/app/_components/sidebar-controller.tsx
@@ -1,3 +1,5 @@
+"use client";
+
 import { usePathname } from "next/navigation";
 import { useEffect } from "react";
 


app/_layouts/shared-layout.tsx
```diff
@@ -79,6 +81,11 @@ export default async function SharedLayout({
   const dictionary = await getDictionary(lang);
   const pageMap = await getPageMap(`/${lang}`);
 
+  // Get the current pathname to determine initial sidebar state
+  const headersList = await headers();
+  const pathname = headersList.get("x-pathname") || "";
+  const isHomePage = pathname.endsWith("/home");
```
evantahler: We also need check on ssr if we are on the homepage or not to start with the sidebar collapsed or not, otherwise you see a jank animation post-pageload.

evantahler:
@sdserranog any idea why the build might be failing here?
```
[Parsing files]
Did not find a data-pagefind-body element on the site.
↳ Indexing all <body> elements on the site.
1 page found without an <html> element.
Pages without an outer <html> element will not be processed by default.
If adding this element is not possible, use the root selector config to target a different root element.
  * "/_not-found.html" has no <html> element

[Reading languages]
Discovered 0 languages:

[Building search indexes]
Total:
  Indexed 0 languages
  Indexed 0 pages
  Indexed 0 words
  Indexed 0 filters
  Indexed 0 sorts
Error: Pagefind was not able to build an index.
Most likely, the directory passed to Pagefind was empty or did not contain any html files.
Error: Pagefind was not able to build an index.
Most likely, the directory passed to Pagefind was empty or did not contain any html files.
 ELIFECYCLE  Command failed with exit code 1.
```

evantahler:
I defer to @vfanelle who was the one who wanted the sidebar hidden.  I'll close this PR either way, because with the layout changes this would have to change dramatically

PR #534, opened 2025-10-24
fixed cli syntax for mcp client config generation

avoguru:


torresmateo (approved):
great catch! thank you!

EricGustin (approved):
Thanks @avoguru !

Alternatively omit `--host local` since local is the default. But never bad to be explicit!

PR #553, opened 2025-11-03
Eric's Documentation Accuracy Review

EricGustin:
I went through all of the pages that relate to the CLI and the server framework under the `en/home` directory. 

I will have a separate PR for all the pages under `en/references/mcp/`

### DIFF
--- a/app/en/home/api-keys/page.mdx
+++ b/app/en/home/api-keys/page.mdx
@@ -64,6 +64,10 @@ Visit the [API Keys page](https://api.arcade.dev/dashboard/api-keys) in Arcade D
 uv tool install arcade-mcp
 ```
 
+<Callout type="info">
+This will install the Arcade CLI as a [uv tool](https://docs.astral.sh/uv/guides/tools/#installing-tools), making it available system wide.
+</Callout>
+
 </Tabs.Tab>
 <Tabs.Tab>
 ```bash
--- a/app/en/home/arcade-cli/page.mdx
+++ b/app/en/home/arcade-cli/page.mdx
@@ -3,7 +3,7 @@ title: "Arcade CLI"
 description: "Learn how to install and use the Arcade CLI"
 ---
 
-import { Tabs } from "nextra/components";
+import { Tabs, Callout } from "nextra/components";
 
 # The Arcade CLI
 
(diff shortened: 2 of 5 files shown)


app/en/home/build-tools/create-a-tool-with-auth/page.mdx
```diff
@@ -341,45 +321,18 @@ async def get_posts_in_subreddit(
...
     # Run the server
-    # - "http" (default): HTTP streaming for Cursor, VS Code, etc.
-    # - "stdio": Standard I/O for Claude Desktop, CLI tools, etc.
+    # - "stdio" (default): Standard I/O transport
+    # - "http": HTTP streamable transport
     app.run(transport=transport, host="127.0.0.1", port=8000)
 
 ```
-
-### Run your MCP server
```
EricGustin: 'Run your MCP server' section is already shown above

app/en/home/build-tools/create-a-tool-with-auth/page.mdx
```diff
@@ -116,12 +118,10 @@ async def get_posts_in_subreddit(
 To use this tool, you need to install the Arcade CLI and run 'arcade login' to authenticate:
 
 ```bash
-uv pip install arcade-mcp
+uv tool install arcade-mcp
 arcade login
 ```
 
-Or add your `ARCADE_API_KEY` and `ARCADE_USER_ID` as environment variables or an `.env` file at the root of your project.
-
```
EricGustin: I want them to login. Also, using env var or .env is finicky when dealing with stdio transport

torresmateo: I agree with this, but I've also found it slightly confusing that if I need a tool with a secret I have to use the `.env` file anyway, especially if I run `arcade configure` where (with stdio):
- I can't get it to reliably inject the Arcade API key
- If I have an .env file, it will load _only_ those variables, and omit the Arcade API key.

So it sound to me that we need to clarify here that the `.env` route is valid (and for secrets maybe required)

EricGustin: I'm not willing to die on this hill haha, so if the following two points don't convince you, then I can revert. Let me know.

1. I'm not convinced that the MCP server developer needs to know that under the hood we use their ARCADE_API_KEY and ARCADE_USER_ID. They just need to know that `arcade login` authenticates them.
2. `arcade login` links their Arcade account with CLI usage & 'MCP server started' events sent to PostHog.

PR #557, opened 2025-11-05
Fix pagefind

evantahler:
Next.js v15.x produced HTML pages on build, and Next.js v16.x does not.  Pagefind (which we use for indexing our site for search) only works with HTML content.  This PR indexes our markdown pages directly stripping out the layout and building a shim DOM to index.


sdserranog (approved):
Hell yeah! :shipit:

PR #734, opened 2026-02-06
Adds metadata alternate links

nearestnabors:
## Summary

Adds a `<link rel="alternate" type="text/markdown">` tag to every page's `<head>`, allowing crawlers and AI agents to discover the markdown version of each page.

**Example:** On `/en/guides/quickstart`, the head contains:
```html
<link rel="alternate" type="text/markdown" href="https://docs.arcade.dev/en/guides/quickstart.md"/>
```

## Why this approach?

The original implementation rendered the link in the root layout's `<Head>` component, but this **didn't update during client-side navigation** - users navigating between pages would still see the original page's `.md` link.

This PR uses Next.js's `generateMetadata` function instead, which:
- Runs on every navigation (including client-side)
- Properly updates the `<head>` metadata
- Generates absolute URLs using `metadataBase`
- Requires no client-side JavaScript

## Changes

- **`proxy.ts` → `middleware.ts`**: Renamed and fixed header passing to use request headers (not response headers) so server components can read the pathname
- **`app/layout.tsx`**: Converted static `metadata` export to async `generateMetadata` function with dynamic `alternates.types`

## Test plan

- [x] Unit tests pass (`pnpm test`)
- [ ] Verify alternate link updates when navigating between pages (no hard refresh needed)
- [ ] Verify crawlers see the correct link in page source

### DIFF
--- a/app/layout.tsx
+++ b/app/layout.tsx
@@ -23,58 +23,6 @@ import {
 
 const REGEX_LOCALE = /^\/([a-z]{2}(?:-[A-Z]{2})?)(?:\/|$)/;
 
-export const metadata = {
-  title: {
-    default: "Arcade Docs",
-    template: "%s | Arcade Docs",
-  },
-  description: "Arcade - AI platform for developers",
-  metadataBase: new URL("https://docs.arcade.dev"),
-  manifest: "/site.webmanifest",
-  icons: {
-    icon: [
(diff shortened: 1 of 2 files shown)


evantahler (approved):
I didn't test this too thoroughly, but let's give it a shot!

PR #738, opened 2026-02-09
Add content negotiation for markdown responses

nearestnabors:
## Summary

- Serve markdown directly when `Accept: text/markdown` header is present
- Auto-detect known AI/LLM agents by User-Agent and serve markdown automatically
- Update `Content-Type` from `text/plain` to `text/markdown` for proper MIME type

This allows AI agents to get markdown in a single request without needing to:
1. Parse HTML to find the `<link rel="alternate">` tag
2. Make a second request to fetch the `.md` file

### Supported AI agents

Claude, ChatGPT, Cursor, Copilot, Gemini, Perplexity, Phind, LangChain, LlamaIndex, Codeium, Tabnine, Amazon Q, and many more.

## Test plan

- [x] Test normal browser request returns HTML
- [x] Test `curl -H "Accept: text/markdown" https://docs-git-feature-content-negotiation-markdown-arcade-ai.vercel.app/en/home` returns markdown
- [x] Test `curl -A "Claude/1.0"https://docs-git-feature-content-negotiation-markdown-arcade-ai.vercel.app/en/home` returns markdown
- [x] Verify `Content-Type: text/markdown` header is returned
- [x] Test that Claude Code returns markdown when pointed athttps://docs-git-feature-content-negotiation-markdown-arcade-ai.vercel.app/en/home
- [x] Test that Claude returns markdown when pointed at https://docs-git-feature-content-negotiation-markdown-arcade-ai.vercel.app/en/home

Failed, but acceptable for now.
- Comet does NOT return markdown, only HTML, when pointed at https://docs-git-feature-content-negotiation-markdown-arcade-ai.vercel.app/en/home


app/api/markdown/[[...slug]]/route.ts
```diff
@@ -34,7 +34,7 @@ export async function GET(
       return new NextResponse(content, {
```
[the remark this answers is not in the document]
Comment (nearestnabors) [reply app/api/markdown/[[...slug]]/route.ts]: Excellent points. Please stand by!

middleware.ts
```diff
@@ -60,9 +118,67 @@ function pathnameIsMissingLocale(pathname: string): boolean {
...
+  if (SUPPORTED_LOCALES.some((loc) => pathname === `/${loc}`)) {
+    return `${pathname}/home.md`;
+  }
+  return `${pathname}.md`;
+}
+
+/**
+ * Handle content negotiation for AI agents and Accept: text/markdown requests
+ */
+function handleContentNegotiation(
```
byrro: I believe trailing slash may produce an invalid markdown path

`/en/home/` becomes `/en/home/.md`, which likely 404s?

we should probably normalize trailing slashes before `buildMarkdownPath`

middleware.ts
```diff
@@ -60,9 +118,67 @@ function pathnameIsMissingLocale(pathname: string): boolean {
...
+  request: NextRequest,
+  pathname: string
+): NextResponse | null {
+  const shouldServeMarkdown = prefersMarkdown(request) || isAIAgent(request);
```
byrro: We should probably limit this to `GET` and `HEAD` requests for the sake of avoiding potential issues with `POST` and `PUT` requests that may be implemented in the future (I believe this wouldn't be an issue today).

Perhaps something like this:

```javascript
  const method = request.method.toUpperCase();
  if (method !== "GET" && method !== "HEAD") {
    return null;
  }
```

middleware.ts
```diff
@@ -5,6 +5,64 @@ const SUPPORTED_LOCALES = ["en", "es", "pt-BR"];
...
+  /glean/i,
+  /devdocs/i,
+  /readable/i,
+];
+
+/**
+ * Check if the request prefers markdown content via Accept header
+ */
+function prefersMarkdown(request: NextRequest): boolean {
+  const acceptHeader = request.headers.get("accept") || "";
```
byrro: We should make this case insensitive, as it would fail if the request comes in with `Accept: Text/Markdown`, for example.

Something like:

```javascript
const acceptHeader = (request.headers.get("accept") || "").toLowerCase();
```

byrro (approved):
LGTM

PR #847, opened 2026-03-03
Some tweaks in tool docs

jottakka:
> [!NOTE]
> **Medium Risk**
> Moderate risk because the docs generator now auto-deletes previously generated toolkit JSON when a toolkit is detected as removed, so any change-detection bug could remove output unexpectedly; other changes are mainly CI/test refactors.
> 
> **Overview**
> **Docs site:** Refactors integration category `_meta.tsx` files to use a new `createCategoryMeta()` helper that auto-inserts **Optimized/Starter** separators based on each entry’s `type`, avoiding duplicated separator boilerplate.
> 
> **CI/tests:** Adds a new `External URL check` GitHub Action running Vitest with `vitest.external-urls.config.ts`, and updates the external URL test to emit warnings (not failures) while excluding it from the default `vitest.config.ts` run.
> 
> **Toolkit docs generator:** Removes stale toolkit JSON entries (`ComplexTools`, `Deepwiki`, `Test2`) from generated data, and extends `generate --skip-unchanged` to collect `removed` toolkits via `collectRemovedToolkitIds()` and pass them through the existing cleanup step so stale `<id>.json` files are deleted and `index.json` is rebuilt (covered by new unit + scenario tests).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b6d0813944da9c84bdf378dac4d4d7c1d1d53750. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


toolkit-docs-generator/excluded-toolkits.txt
```diff
@@ -1,3 +1,8 @@
 # Toolkits to exclude from docs generation
 Google
 Microsoft
+ComplexTools
+Deepwiki
+Test2
+MyServer
+Mytest
```
EricGustin: Why were the Test2, DeepWiki, etc. toolkits being pulled from the engine in the first place? Adding these to an exclude list feels like a hack instead of address the root cause, but I could be missing something

EricGustin: So does that mean the pipeline won't automatically remove toolkits that are no longer hosted by the Engine?

PR #907, opened 2026-04-09
feat(toolkit-docs-generator): latest-version coherence filter

jottakka:
## Summary

- Adds `filterToolsByMajorityVersion()` utility that computes the version shared by the most tools in a toolkit and drops tools at other versions
- Applied in both `fetchAllToolkitsData()` and `fetchToolkitData()` (when no explicit version is passed) in `CombinedToolkitDataSource`
- Fixes stale tools from older toolkit releases (e.g. Github notification tools at `@2.0.1` alongside 35+ tools at `@3.1.3`) surviving through the docs pipeline indefinitely

## Problem

The Engine `/v1/tool_metadata` endpoint returns tools at mixed versions for the same toolkit. The docs generator trusts this as-is, grouping tools by toolkit ID with no version filtering. The `--skip-unchanged` flow compounds this — it correctly sees no change run after run (because the API keeps returning the same stale tools), preserving old output indefinitely.

## Solution

After grouping tools by toolkit ID, compute the majority version (the version shared by the most tools), then keep only tools at that version. This drops stale tools before they reach the diff or merger. Explicit version parameters bypass the filter.

## Test plan

- [x] Unit tests for `getMajorityVersion` and `filterToolsByMajorityVersion` (11 tests)
- [x] Integration tests for `CombinedToolkitDataSource` version filtering (3 new tests)
- [x] Scenario test: mixed-version tools → filter → clean diff (3 tests)
- [x] Full test suite passes (499 tests, 0 type errors)

---

> [!NOTE]
> **Medium Risk**
> Changes which tools are included in generated docs/diffs by automatically dropping older-version tools when Engine returns mixed versions, which can cause large output churn if current data has inconsistencies.
> 
> **Overview**
> Adds a highest-version coherence filter so each toolkit keeps only tools at the numerically highest `@version` when no explicit version is requested, dropping stale tools from older releases in both `fetchToolkitData()` and `fetchAllToolkitsData()`.
> 
> Introduces `utils/version-coherence.ts` (plus exports) and moves `extractVersion` into `utils/fp.ts` while re-exporting it from `data-merger.ts` for compatibility. Updates the merger to **preserve a previous toolkit summary** when the tool signature changes but summary generation is disabled (no LLM), and adds unit/scenario tests covering mixed-version filtering and diff/summary behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 417a51fd98dfba74227b471639b2f5e76a54b215. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/toolkit-docs-generator/src/sources/toolkit-data-source.ts
+++ b/toolkit-docs-generator/src/sources/toolkit-data-source.ts
@@ -9,7 +9,7 @@
 import { join } from "path";
 import type { ToolDefinition, ToolkitMetadata } from "../types/index.js";
 import { normalizeId } from "../utils/fp.js";
-import { filterToolsByMajorityVersion } from "../utils/version-coherence.js";
+import { filterToolsByHighestVersion } from "../utils/version-coherence.js";
 import {
   type ArcadeApiSourceConfig,
   createArcadeApiSource,
@@ -131,14 +131,14 @@ export class CombinedToolkitDataSource implements IToolkitDataSource {
       }
     }
 
-    // Filter tools by version if specified, otherwise apply majority-version
-    // coherence to drop stale tools from older releases that Engine still serves.
+    // Filter tools by version if specified, otherwise keep only the highest
+    // version to drop stale tools from older releases that Engine still serves.
     const filteredTools = version
       ? tools.filter((tool) => {
           const toolVersion = tool.fullyQualifiedName.split("@")[1];
           return toolVersion === version;
         })
-      : filterToolsByMajorityVersion(tools);
+      : filterToolsByHighestVersion(tools);
 
     return {
       tools: filteredTools,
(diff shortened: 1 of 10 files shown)


EricGustin:
wouldn't we want to continue documenting the deprecated tools (but with a deprecated banner) since the engine is still hosting them?

EricGustin (approved):

toolkit-docs-generator/tests/utils/version-coherence.test.ts
teallarson, line 1: I don't see a test here for comparing a release candidate against a version with teh same number (ie: `1.23-rc.1` vs `1.23`, and I _think_ based on the logic above where we strip out everything after `-` those would evaluate as equal.

Is that a scenario we should guard against?

toolkit-docs-generator/src/sources/toolkit-data-source.ts
teallarson, line 1: it's not a new issue in this PR, but i did notice that we do a deeper hunt for toolkit data when looking up one toolkit than we do when listing... which means the listed toolkit's data (icon, url, etc.) might differ from a single fetch toolkit's data.

teallarson (commented):
I _think_ we're still missing an edge case around pre-release versions.

PR #921, opened 2026-04-17
fix: follow up on #907 version edge cases

jottakka:
## Summary
- make version coherence semver-aware so stable releases outrank matching pre-release versions (for example, `1.23` over `1.23-rc.1`)
- add regression tests for stable-vs-rc filtering behavior in `version-coherence`
- reuse the same providerId metadata fallback in `fetchAllToolkitsData` as `fetchToolkitData` and add a consistency test for `MailchimpMarketingApi`
- address reviewer concern about RC-vs-stable precedence by adding semver ordering tests (`rc.10 > rc.2`, numeric vs string identifiers, and prefix-length precedence)
- address bugbot’s redundant fetch comment by preventing a second `getToolkitMetadata(toolkitId)` call in `fetchToolkitData`
- add call-count regression tests proving single fetch does not re-fetch metadata while bulk fetch still performs toolkit-id lookup on metadata-map misses

## Test plan
- [x] `pnpm vitest run toolkit-docs-generator/tests/utils/version-coherence.test.ts toolkit-docs-generator/tests/sources/toolkit-data-source.test.ts`

Made with [Cursor](https://cursor.com)

---

> [!NOTE]
> **Medium Risk**
> Changes toolkit version-selection logic (SemVer prerelease/build precedence) and metadata resolution fallback, which can alter which tools/metadata are emitted in generated docs. Covered by new regression tests, but affects core selection behavior across single and bulk fetch paths.
> 
> **Overview**
> Makes version coherence *SemVer-precedence aware* by introducing a proper `compareVersions` that orders stable releases above matching pre-releases, compares prerelease identifiers correctly (numeric vs string, prefix length), and treats build metadata as ordering-ignored but equivalence-preserving. `filterToolsByHighestVersion` now keeps all tools semver-equivalent to the highest release (e.g. `1.0.0` and `1.0.0+build.1`) instead of string-matching only.
> 
> Unifies and reuses the providerId-based metadata fallback for `*Api` toolkits via `resolveProviderMetadata`, applying it consistently in both `fetchToolkitData` and `fetchAllToolkitsData`, while avoiding redundant `getToolkitMetadata(toolkitId)` calls in the single-toolkit path and only doing per-toolkit lookups in bulk when the batch metadata map misses.
> 
> Adds targeted tests covering SemVer edge cases (stable vs rc, rc ordering, ASCII ordering, build metadata equivalence) and metadata lookup consistency/call-count behavior for both single and bulk fetches.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5528dcd947cbc9938eab04dfff5fa261d15b93ec. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/toolkit-docs-generator/src/sources/toolkit-data-source.ts
+++ b/toolkit-docs-generator/src/sources/toolkit-data-source.ts
@@ -108,30 +108,29 @@ export class CombinedToolkitDataSource implements IToolkitDataSource {
     this.metadataSource = config.metadataSource;
   }
 
-  private async resolveMetadata(
+  /**
+   * Apply the "*Api" provider-id fallback when the direct metadata lookup
+   * missed. Used by both `fetchToolkitData` and `fetchAllToolkitsData` so
+   * they resolve metadata the same way.
+   */
+  private async resolveProviderMetadata(
     toolkitId: string,
     tools: readonly ToolDefinition[],
-    directMetadata: ToolkitMetadata | null,
-    options?: { readonly allowToolkitIdLookup?: boolean }
+    directMetadata: ToolkitMetadata | null
   ): Promise<ToolkitMetadata | null> {
-    let metadata = directMetadata;
-    if (!metadata && options?.allowToolkitIdLookup) {
-      metadata = await this.metadataSource.getToolkitMetadata(toolkitId);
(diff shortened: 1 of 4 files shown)


teallarson (approved):
that should do it

PR #927, opened 2026-04-17
fix(toolkit-docs-generator): preserve previous summary when regen fails

### LINKED ISSUE #926: Toolkit docs generator silently wipes summary on LLM unavailable / failure
## Summary

`DataMerger.maybeGenerateSummary` in [toolkit-docs-generator/src/merger/data-merger.ts](https://github.com/ArcadeAI/docs/blob/main/toolkit-docs-generator/src/merger/data-merger.ts) silently drops the toolkit `summary` field in two paths:

1. **No LLM generator**: when `toolkitSummaryGenerator` is undefined (e.g. `--skip-summary`, or missing LLM env vars on a scheduled run), and the tool signature has changed since the last output (so the reuse fast path doesn't match). The function returns without copying the previous summary forward.
2. **LLM call fails**: when the generator throws (rate limit, transient API error, invalid JSON response), the `catch` block records a warning but leaves `result.toolkit.summary` undefined.

Because the rendered toolkit page in `app/_components/toolkit-docs/components/toolkit-page.tsx` only shows the prose block when `data.summary` is truthy, this regression manifests on production Vercel builds as the overview block disappearing entirely for affected toolkits.

## Impact

Rich, hand-refined summaries for the following toolkits were wiped to `null` across recent `[AUTO] Adding MCP Servers docs update` merges:

- `github` (wiped in PR #922)
- `jira` (wiped in PR #922)
- `googledocs` (wiped in PR #915)
- `linear` (wiped in PR #915)
- `salesforce` (wiped in PR #899)
- `daytona` (wiped in PR #899)

## Root cause

Reuse fast path only matches when `buildToolkitSummarySignature` produces the same output — any tool add/remove/parameter change invalidates it. From there, both "no generator" and "generator threw" paths fall through to a no-op instead of falling back to the previous summary.

An earlier fix for path 1 was written on the PR #907 feature branch (commit `417a51fd fix: preserve previous summary when no LLM is available to regenerate`) but was dropped during the merge to main.

## Fix (proposed)

In both fallback paths, carry `previousToolkit?.summary` forward onto `result.toolkit.summary`. A slightly stale summary is strictly better than losing hand-refined content on every affected run.

Follow-up PRs will restore / draft truthful summaries for the six toolkits above plus the five that never received one (`posthog`, `granola`, `microsoftoutlookcalendar`, `microsoftoutlookmail`, `zohocreatorapi`).

## Repro

1. Run the docs generator for a toolkit that has a non-null summary in the previous output.
2. Add or remove a tool so `buildToolkitSummarySignature` changes.
3. Either pass `--skip-summary` / unset the LLM env, or stub the LLM to throw.
4. Observe: output JSON has `summary: null` (was populated).

## Linked PR

- Fix: `fix/toolkit-docs-preserve-summary` (draft)
- Restoration: separate draft PRs, linked from the fix PR.

jottakka:
## Summary

- `maybeGenerateSummary` silently wiped the toolkit `summary` field in two fallback paths — when no LLM generator was configured and when the LLM call threw.
- Both paths now carry the previous toolkit's `summary` forward, so a signature change never loses hand-refined content on its own.
- Adds two regression tests: missing-generator + failing-generator.

## Why

Rendered toolkit pages only show the overview prose block when `data.summary` is truthy ([app/_components/toolkit-docs/components/toolkit-page.tsx](https://github.com/ArcadeAI/docs/blob/main/app/_components/toolkit-docs/components/toolkit-page.tsx#L647)). The silent wipe meant that every \`[AUTO] Adding MCP Servers docs update\` merge that touched a toolkit's tool set could drop its overview on Vercel with no warning surfaced to the reader.

An earlier version of this fix (missing-generator path only) was written on the PR #907 branch but got dropped during merge to `main`. This PR restores it and additionally covers the LLM-failure path.

## Scope

Changes are isolated to `toolkit-docs-generator/src/merger/data-merger.ts` (+ tests). No generator output regenerated in this PR — summary restoration is split into follow-up PRs so reviewers can evaluate content separately from the code fix.

## Test plan

- [x] \`pnpm vitest run toolkit-docs-generator/tests/merger/data-merger.test.ts\` — 58 passed (+2 new)
- [ ] Confirm CI green

Closes #926

---

> [!NOTE]
> **Medium Risk**
> Changes toolkit JSON output semantics by preserving prior `summary` on LLM-unavailable/failure paths and introducing `summaryStale` flags; downstream consumers/CI may start failing if stale summaries are produced.
> 
> **Overview**
> Prevents toolkit overviews from being silently dropped by **carrying forward the previous `summary`** when summary regeneration is skipped (no LLM configured) or fails, instead of wiping the field.
> 
> Introduces `summaryStale`/`summaryStaleReason` on `MergedToolkit` and propagates these through previous-output parsing; summaries reused via signature match are now only considered fresh if the prior summary was not already flagged stale.
> 
> Adds a CI gate (`tests/stale-summaries.test.ts`) and a companion script (`scripts/check-stale-summaries.ts`) to fail/report when committed toolkit JSON contains stale summaries, plus expanded `DataMerger` tests covering stale-flag behavior and clearing on successful regeneration.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f5a58e6c28e833a15debe67a57e9abe0827c49a6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/toolkit-docs-generator/scripts/check-stale-summaries.ts
+++ b/toolkit-docs-generator/scripts/check-stale-summaries.ts
@@ -26,16 +26,32 @@ type ToolkitShape = {
   summaryStaleReason?: unknown;
 };
 
-const listToolkitFiles = (): string[] =>
-  readdirSync(TOOLKITS_DIR).filter(
-    (name) => name.endsWith(".json") && name !== "index.json"
-  );
+const listToolkitFiles = (): string[] => {
+  try {
+    return readdirSync(TOOLKITS_DIR).filter(
+      (name) => name.endsWith(".json") && name !== "index.json"
+    );
+  } catch (error) {
+    const message = error instanceof Error ? error.message : String(error);
+    console.error(
+      `✗ Cannot read toolkit directory ${TOOLKITS_DIR}: ${message}`
+    );
+    process.exit(1);
+  }
+};
 
 const main = (): number => {
   const stale: Array<{ file: string; id: string; reason: string }> = [];
   for (const file of listToolkitFiles()) {
-    const raw = readFileSync(join(TOOLKITS_DIR, file), "utf8");
-    const toolkit = JSON.parse(raw) as ToolkitShape;
+    let toolkit: ToolkitShape;
+    try {
+      const raw = readFileSync(join(TOOLKITS_DIR, file), "utf8");
+      toolkit = JSON.parse(raw) as ToolkitShape;
+    } catch (error) {
+      const message = error instanceof Error ? error.message : String(error);
+      console.error(`✗ Cannot parse ${file}: ${message}`);
+      return 1;
+    }
     if (toolkit.summaryStale === true) {
       stale.push({
         file,
--- a/toolkit-docs-generator/scripts/check-stale-summaries.ts
+++ b/toolkit-docs-generator/scripts/check-stale-summaries.ts
@@ -1,0 +1,66 @@
+#!/usr/bin/env tsx
+
+/**
+ * Check stale summaries
+ *
+ * Scans toolkit-docs-generator/data/toolkits/*.json and reports any toolkit
+ * whose summary was carried forward from a previous run (summaryStale: true)
+ * because regeneration was skipped or failed. Exits 1 if any are found so
+ * the script can gate CI or a local sanity check.
+ *
+ * Usage:
+ *   pnpm dlx tsx toolkit-docs-generator/scripts/check-stale-summaries.ts
+ */
+
+import { readdirSync, readFileSync } from "node:fs";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
+
+const here = dirname(fileURLToPath(import.meta.url));
+const TOOLKITS_DIR = join(here, "..", "data", "toolkits");
+
+type ToolkitShape = {
+  id?: unknown;
+  summary?: unknown;
+  summaryStale?: unknown;
+  summaryStaleReason?: unknown;
+};
+
+const listToolkitFiles = (): string[] =>
+  readdirSync(TOOLKITS_DIR).filter(
+    (name) => name.endsWith(".json") && name !== "index.json"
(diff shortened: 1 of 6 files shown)


jottakka:
## Follow-up: stale-summary flag + CI gate

Added in commit 9e392f57 — scoped to this preservation PR as requested; the deterministic scans/enforcement live in #932.

**Shape:**
- `MergedToolkit` gets optional `summaryStale: boolean` + `summaryStaleReason: string`.
- Set together whenever `maybeGenerateSummary` carries the previous summary forward because regen was skipped (no LLM) or threw. A warning (\`Summary is stale for <id>: <reason>. Previous summary carried forward.\`) is also pushed onto the merge result.
- Cleared whenever a fresh summary is generated, the signature-match reuse path confirms the existing summary is still accurate, or an overview chunk takes over.

**CI gate:**
- [\`tests/stale-summaries.test.ts\`](https://github.com/ArcadeAI/docs/blob/fix/toolkit-docs-preserve-summary/toolkit-docs-generator/tests/stale-summaries.test.ts) scans committed \`data/toolkits/*.json\` and fails if any toolkit has \`summaryStale: true\`, with a list of offenders + reasons. The scheduled [AUTO] PR still gets created; CI fails on it — which is the whole point.
- [\`scripts/check-stale-summaries.ts\`](https://github.com/ArcadeAI/docs/blob/fix/toolkit-docs-preserve-summary/toolkit-docs-generator/scripts/check-stale-summaries.ts) is the same scan as a local CLI.

**Tests added:**
- Existing preservation tests now assert `summaryStale`/reason are set.
- New test: successful regen clears an existing stale flag.
- New test: signature-match reuse clears an existing stale flag.

526 tests passing, type-check clean.

byrro (approved):

PR #947, opened 2026-04-24
feat(integrations): URL-anchored filters

vfanelle:
Preview: https://docs-git-vfanelle-integrations-url-anchored-filters-arcade-ai.vercel.app/en/resources/glossary#pro-tool-executions

URL query params are now the source of truth for the integrations page filters. A filtered view can be copied from the address bar and shared — paste the URL anywhere and the recipient sees the same filtered list of MCP servers. The glossary's Pro Tool Executions and BYOC sections link directly to these views.

## Link structures

| What you want to see | URL |
|---|---|
| Pro tools only | `/en/resources/integrations?label=pro#list` |
| BYOC-eligible tools only | `/en/resources/integrations?label=byoc#list` |
| Tools that are both Pro and BYOC | `/en/resources/integrations?label=pro&label=byoc#list` |
| All MCP servers (no filters) | `/en/resources/integrations` |

The `#list` anchor scrolls past the page header straight to the tool list.

Made with [Cursor](https://cursor.com)


teallarson (changes requested):
I love the move to search params! This can be simpler, though. 

The main thing I’d avoid is having two sources of truth: Zustand filter state and URL params, then syncing them through a global store subscription, mount hydration, replaceState, and a manual popstate listener. Could we make the URL/search params the source of truth instead, and have filter handlers write params directly? That should remove most of useUrlFilterSync. For read, use useSearchParams()! For writing, since the filtering is entirely client-side and doesn’t need a server navigation/refetch, the native History API seems reasonable. Next documents that pushState / replaceState integrate with useSearchParams and usePathname.

A few specific edge cases in the current version:

- The PR says URL params are applied before first paint, but this uses useEffect, so they apply after paint/hydration.
- Filter changes use replaceState, so back/forward will not step through filter changes. If that behavior matters, this should use pushState; if it doesn’t, the PR description/test plan should not claim it.
- Landing on /integrations with no query params does not reset a previously populated global Zustand store.
category is not validated, so ?category=bogus can create invalid filter state.
- Boolean parsing is loose: ?pro=bogus currently counts as truthy, even though malformed params are meant to be ignored.

There are also some typing issues that point in the same direction. VALID_TYPES is a Set<string>, then the parser recovers with as ToolkitType; and SerializableFilterState.selectedType is just string, so invalid values can be serialized without TypeScript noticing. I’d rather share the actual filter value shape, e.g. selectedType: ToolkitType | "all", use a real type guard for toolkit types, and validate categories if category is URL-addressable.

Minor test-structure nit: the tests import pure parse/serialize helpers from a "use client" hook module. It may be cleaner to split those helpers into a non-client module and keep the hook/component layer separate.

app/en/resources/integrations/components/filter-params.ts
teallarson, line 1: There's a risk here that if the design system removes a category this will break, or if the design system adds a category, it won't get added here. 

We could use: `const TOOLKIT_TYPES = Object.keys(TYPE_CONFIG) as ToolkitType[];` and const TOOLKIT_CATEGORIES = CATEGORIES.map((c) => c.id) as ToolkitCategory[];

I don't know that it's worth blocking on as I don't think these change often.

teallarson (approved):
Tested by visiting https://docs-git-vfanelle-integrations-url-anchored-filters-arcade-ai.vercel.app/en/resources/integrations?label=pro#list 🎉

PR #1007, opened 2026-06-11
fix(seo): clear remaining docs audit errors (MARTECH-19)

juan-arcadedev:
## What

Follow-up to [MARTECH-16](https://linear.app/arcadedev/issue/MARTECH-16) that clears the remaining Ahrefs Site Audit errors on docs.arcade.dev: "links to broken page", 404/4XX, "duplicate pages without canonical", and "broken redirect". Page-size errors are out of scope ([MARTECH-17](https://linear.app/arcadedev/issue/MARTECH-17)).

Linear: [MARTECH-19](https://linear.app/arcadedev/issue/MARTECH-19)

## Changes

**1. `/en/resources` 404 (~115 incoming links — highest impact)**
The toolkit breadcrumb links "Resources" → `/en/resources`, which has no page. Added a permanent redirect → `/en/resources/integrations`. Breadcrumb left as-is.

**2 & 3. Integrations index linked to non-existent pages + duplicate Notion**
Root cause: the design-system catalog carries bare-name entries (e.g. `Datadog` alongside `DatadogApi`) and doc-less placeholders that have **no generated page** — the index rendered a clickable card for each → 404.
- New `resolveIndexToolkits` (`app/_lib/integration-index.ts`) + `listValidIntegrationLinks` (`app/_lib/toolkit-static-params.ts`) resolve each card against the real route set: dynamic `[toolkitId]` routes **∪** authored static pages (e.g. partner pages, which are static MDX files, not `[toolkitId]` routes).
- Bare `-api` duplicates are dropped (the real card stays); same-URL entries (e.g. Notion ×2) are de-duped; toolkits with no page stay visible but **non-clickable** (existing "coming soon" `<button>` path — no href emitted).
- Added a self-referential `canonical` to every toolkit page.
- `getToolkitsWithDocsLinks` extracted to `app/_lib/integration-catalog.ts`.

The resolver is **data-driven**: a card is clickable iff its page actually exists, so it adapts automatically as toolkits gain or lose pages (see "Kept current with main").

**4. "Broken redirect" (squareup)**
Was actually a broken link — `square/page.mdx` pointed at `…/productivity/squareupapi`; fixed to `squareup-api`.

**5. In-content emails → `/cdn-cgi/l/email-protection`**
> ⚠️ The ticket's `<!--email_off-->` marker is **infeasible in this repo**: raw HTML comments hard-error the Nextra/MDX build (`Unexpected character !`), and the only React way to emit one (`dangerouslySetInnerHTML`) is a banned Biome error (`lint/security/noDangerouslySetInnerHtml`; suppressions are disallowed by CLAUDE.md).

Instead, new `<ContactEmail>` (`app/_components/contact-email.tsx`) assembles the `mailto:` only after mount; SSR/crawlers see a plain `/en/resources/contact-us` link, so Cloudflare has nothing to obfuscate (also aligns with the site's anti-scraping posture). Applied to faq, auth-providers, oauth2, and security-research-program (×2). `chrome-no-mailto.test.ts` exempts this one sanctioned helper.

## Regression guard

`tests/integration-index-links.test.ts` covers the class of links the MDX-only `broken-link-check` test can't see (component-generated links + dynamic `[toolkitId]` slugs). It's intentionally **data-independent** so toolkit-data changes don't make it stale (an early version hard-coded `datadog` and broke once `main` gave it a real page):
- A **unit test** of `resolveIndexToolkits` using placeholder toolkits — deterministically covers drop / de-dup / hidden / non-clickable, with no dependency on real data.
- **Live-data invariants** derived from the catalog + route set: no clickable card points at a missing page, links are unique (dedup), the index renders ≥1 real card, and authored static pages are recognized.
- Validates the breadcrumb's hardcoded internal hrefs against real pages / routes / redirects.

Required adding a `@/` alias to `vitest.config.ts` (no test imported app modules before).

### DIFF
--- a/app/_components/contact-email.tsx
+++ b/app/_components/contact-email.tsx
@@ -4,6 +4,10 @@ import Link from "next/link";
 import type { ReactNode } from "react";
 import { useEffect, useState } from "react";
 
+// TODO(i18n): hardcoded to the English contact page. Docs content is en-only
+// today, so this is the safe target (a /<locale>/ link could 404 for locales
+// without the page). When translated content ships, derive the active locale
+// (e.g. via usePathname) and point at /<locale>/resources/contact-us.
 const CONTACT_PAGE = "/en/resources/contact-us";
 
 type ContactEmailProps = {
--- a/app/_components/contact-email.tsx
+++ b/app/_components/contact-email.tsx
@@ -1,0 +1,41 @@
+"use client";
+
+import Link from "next/link";
+import type { ReactNode } from "react";
+import { useEffect, useState } from "react";
+
+const CONTACT_PAGE = "/en/resources/contact-us";
+
+type ContactEmailProps = {
+  /** Local part of the address, e.g. "security" for security@arcade.dev. */
+  user: string;
+  /** Domain of the address, e.g. "arcade.dev". */
+  domain: string;
+  /** Visible link text. Keep it free of the raw address (see below). */
+  children: ReactNode;
+};
+
+/**
+ * A mailto link whose address is assembled only after hydration.
+ *
+ * Cloudflare's Email Address Obfuscation runs at the edge on the server-rendered
+ * HTML and rewrites any `mailto:`/address it finds into a
+ * `/cdn-cgi/l/email-protection` link, which crawlers report as broken. By keeping
+ * the address out of the SSR markup — we render a plain link to the contact page
+ * until the component mounts — there is nothing for Cloudflare to rewrite, while
+ * real users still get a working `mailto:`. Pass visible text via `children`
+ * (not the literal address) so it isn't exposed server-side either.
+ */
+export function ContactEmail({ user, domain, children }: ContactEmailProps) {
+  const [address, setAddress] = useState<string | null>(null);
+
+  useEffect(() => {
+    setAddress(`${user}@${domain}`);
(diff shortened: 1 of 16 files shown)


app/_components/contact-email.tsx
```diff
@@ -0,0 +1,41 @@
+"use client";
+
+import Link from "next/link";
+import type { ReactNode } from "react";
+import { useEffect, useState } from "react";
+
+const CONTACT_PAGE = "/en/resources/contact-us";
```
sdserranog, line 11: This is a clever solution!
nit: this is correct now, but in the future, if we add translations, this would always redirect to the English version, which is ok for now, as I said, but maybe add a TODO or a comment here.

app/_lib/integration-catalog.ts
```diff
@@ -0,0 +1,52 @@
+import type { Toolkit } from "@arcadeai/design-system";
```
sdserranog, line 1: Thanks for extracting these

sdserranog (approved):
:shipit:

app/_components/contact-email.tsx
juan-arcadedev: Good call! 
I'll add a TODO here.

PR #1009, opened 2026-06-12
fix(deps): upgrade @arcadeai/design-system to 3.45.3

sdserranog:
## What

Upgrades `@arcadeai/design-system` from **3.45.2 → 3.45.3** and removes the temporary `pnpm` patch.

## Why

3.45.0–3.45.2 were built with Vite 8 (which swapped Rollup for **Rolldown**). For a bundled CJS dependency (React's `use-sync-external-store/shim`, pulled in transitively by Base UI popover/menu components), Rolldown emitted a runtime `require("react")` shim instead of an ESM import. That shim:

- **throws in the browser** (`Uncaught Error: Calling require for "react" in an environment that doesn't expose the require function`), and
- **resolves React to `null` during SSR** → `TypeError: Cannot read properties of null (reading 'useSyncExternalStore')` → HTTP 500.

The result was that every docs page using those components crashed — navigation broke on all pages except the landing page. `pnpm build` passed regardless (both are runtime failures), so it shipped silently.

3.45.3 fixes the Rolldown external handling upstream (via rolldown's `esmExternalRequirePlugin`), so the published `dist` no longer contains the require shim. This lets us run the latest version with **no consumer-side workaround**.

## Verification

- `pnpm build` exits 0
- Installed `dist` and emitted client chunks contain **zero** `rolldown.rs` require-shim markers
- SSR returns **200** on previously-broken pages (`/en/get-started/about-arcade`, `/en/references/changelog`, `/en/resources/faq`); home still 200
- Browser console is clean on the affected pages (no `require`/React error)

---

> [!NOTE]
> **Low Risk**
> Dependency-only bump with no app source changes; risk is limited to possible UI regressions from the design-system patch release.
> 
> **Overview**
> Bumps **`@arcadeai/design-system`** from `3.45.2` to **`3.45.3`** in `package.json` and refreshes **`pnpm-lock.yaml`** to match (including updated integrity and transitive **`@floating-ui/utils`** pins under Floating UI).
> 
> This patch release is meant to pick up upstream fixes for Rolldown-built `dist` output that previously injected browser/SSR-breaking `require("react")` shims in design-system UI (e.g. Base UI popover/menu paths), so docs can run the latest package without consumer workarounds.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ea0b028c2558ff6f8f3455eec02050b5b972de8d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
   },
   "homepage": "https://arcade.dev/",
   "dependencies": {
-    "@arcadeai/design-system": "3.45.2",
+    "@arcadeai/design-system": "3.45.3",
     "@mdx-js/mdx": "3.1.1",
     "@mdx-js/react": "3.1.1",
     "@next/third-parties": "16.1.7",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -13,8 +13,8 @@ importers:
   .:
     dependencies:
       '@arcadeai/design-system':
-        specifier: 3.45.2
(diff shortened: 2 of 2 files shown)


teallarson (approved):
visiting pages works again 🎉

PR #1023, opened 2026-06-16
fix(seo): clear remaining docs Ahrefs errors — page size, email obfuscation, canonical orphans (MARTECH-17)

juan-arcadedev:
Clears the three remaining **docs.arcade.dev** Ahrefs Site Audit error classes (the 36 error URLs at health 95), per [MARTECH-17](https://linear.app/arcadedev/issue/MARTECH-17). Decisions this round: keep Cloudflare Email Obfuscation **on** + fix in-repo; shrink pages via **lazy-loaded detail with no IA/URL change** (no pagination).

## 1. Page size > 2 MB (20 pages)
Toolkit reference pages serialized the entire `ToolkitData` into the initial payload and server-rendered every tool. Now:
- `toToolkitSummary()` strips the heavy per-tool fields (`parameters`/`output`/`codeExample`) from the client `ToolkitPage` payload.
- Per-tool detail is **lazy-fetched on expand** from the existing `/api/toolkit-data/[toolkitId]` route (new `useToolDetail` hook, one cached fetch per page).
- The per-tool sections and the scope picker render **client-side after mount**, so the server HTML carries only the crawlable summary (Available Tools table + names-only sidebar). The per-link secret icon (793 inline SVGs ≈ 404 KB on github-api) is dropped from the sidebar — secret info stays in the table's Secrets column and the expanded tool detail.

**Result (measured on a prod build): github-api 10.3 MB → 1.40 MB; posthog 1.50 MB; all 20 pages now < 2 MB.** Verified in a browser: click "Show details" lazy-loads + renders the parameters table; deep-links auto-expand + scroll; sidebar search still filters.

## 2. Cloudflare email obfuscation (~16 URLs)
Cloudflare rewrites email-like text into `/cdn-cgi/l/email-protection` (a 404 for crawlers). New `neutralize-emails.tsx` inserts a zero-width `<wbr>` before each `@` (invisible, copy-safe) in toolkit summaries + tool descriptions, so there's no contiguous match. 5 guide example emails reworded to placeholders. (Cloudflare ignores the `<script>` RSC payload and JSON API responses, so the lazy-loaded detail is safe automatically.)

## 3. Canonical orphans (2)
`generateMetadata` now canonicalizes to the toolkit's **own** category + slug (new `getToolkitCanonicalPath`), so the wrong-category alias `development/pagerduty-api` (a docsLink/category data mismatch) points at the linked `customer-support/pagerduty-api` instead of self. Hidden toolkits (`notion`) emit `robots: noindex`.

## Tests
- `tests/page-size.test.ts` — every toolkit's summary stays under budget and keeps no heavy fields.
- `tests/neutralize-emails.test.tsx` — the `<wbr>` neutralizer leaves no contiguous email in rendered output.
- Extended `tests/integration-index-links.test.ts` — no non-hidden toolkit canonicalizes to an orphan.


app/_components/toolkit-docs/components/tool-section.tsx
```diff
@@ -475,38 +554,70 @@ export function ToolSection({
...
-        hasScopes={hasScopes}
-        hasSecrets={hasSecrets}
-        isSelected={isSelected}
-        onToggleSelection={onToggleSelection}
-        showSelection={showSelection}
-        tool={tool}
-      />
-      <ToolMetadataSection metadata={tool.metadata} />
-      <ToolDescriptionSection showDescription={showDescription} tool={tool} />
-      <ToolParametersSection showParameters={showParameters} tool={tool} />
```
sdserranog, line 490: Removing parameters/output here triggers ScopePicker's fallback branch for every toolkit. The copied "Selected tools" JSON loses parameters/output and also changes the name from the qualified name to the short name, which breaks downstream tool configs that expect the qualified name.

Suggestion: have "Copy selected" call loadToolkitDetail(toolkitId) and rebuild the full JSON on demand. If the detail isn't loaded yet, fall back, but emit the name: t.qualifiedName ?? t.name so the identifier stays correct.

app/_components/toolkit-docs/components/toolkit-page.tsx
```diff
@@ -522,6 +517,25 @@ export function ToolkitPage({ data }: ToolkitPageProps) {
...
+    update();
+    window.addEventListener("hashchange", update);
+    return () => window.removeEventListener("hashchange", update);
+  }, []);
+
+  // The per-tool sections render after mount (client-only). The crawlable
+  // summary in the server HTML is the Available Tools table + the sidebar; for
+  // large toolkits (e.g. github-api, 818 tools) server-rendering every section
+  // would blow past Googlebot's 2 MB uncompressed-HTML crawl limit.
+  const [sectionsMounted, setSectionsMounted] = useState(false);
```
sdserranog: **Problem**: This is a manual mount flag that causes an extra re-render to achieve client-only rendering.
**Fix**: wrap the sections block in next/dynamic(() => ..., { ssr: false }) and delete the flag.
https://nextjs.org/docs/pages/guides/lazy-loading#with-no-ssr

app/_components/toolkit-docs/components/use-toolkit-detail.ts
```diff
@@ -0,0 +1,99 @@
...
+  | { status: "error" }
+  | { status: "ready"; tool: ToolDefinition };
+
+export function useToolDetail(
+  toolkitId: string,
+  qualifiedName: string,
+  enabled: boolean,
+  reloadToken = 0
+): ToolDetailState {
+  const [state, setState] = useState<ToolDetailState>({ status: "loading" });
```
sdserranog: **Problem**: The hook reports "loading" even when enabled is false, so the state lies about what it’s doing.
**Suggestion**: start in an "idle" state and switch to "loading" only when enabled is true.

app/en/resources/integrations/_lib/toolkit-docs-page.tsx
```diff
@@ -68,7 +78,9 @@ export function createToolkitDocsPage(category: IntegrationCategory) {
       notFound();
     }
 
-    return <ToolkitPage data={data} />;
+    // Pass a summary (per-tool detail stripped) so the heavy fields never enter
+    // the initial Flight payload — detail is fetched on expand. See MARTECH-17.
+    return <ToolkitPage data={toToolkitSummary(data)} />;
```
sdserranog, line 83: **nit**: the agent markdown view is generated at the edge from the rendered HTML (no in-repo markdown route, copy-page-override.tsx just refetches with Accept: text/markdown). Stripping tool detail from SSR and making sections client-only means that markdown now loses parameters, output, and examples for toolkit pages. Classic SEO and llms.txt are unaffected.

**Suggestion**: add a server markdown route for these pages that builds markdown directly from the full ToolkitData, independent of the slimmed HTML.

app/_components/toolkit-docs/components/tool-section.tsx
juan-arcadedev: good catch — fixed in 2d495509. copy selected now lazy-fetches the full per-tool detail and keeps the qualified name, instead of dropping to the short-name fallback.

app/_components/toolkit-docs/components/toolkit-page.tsx
juan-arcadedev: addressed in 2d495509 — swapped the mount flag for next/dynamic({ ssr: false }) on the sections + scope picker.

app/_components/toolkit-docs/components/use-toolkit-detail.ts
juan-arcadedev: addressed in 2d495509 — the hook starts idle now and only flips to loading once the section is enabled.

app/en/resources/integrations/_lib/toolkit-docs-page.tsx
juan-arcadedev: addressed in 85e71538: added a data-derived markdown route (content-negotiated on /api/toolkit-data) and pointed copy-page-override at it for toolkit pages, so the agent/markdown view keeps params/output/examples. heads up: hitting the page URL directly with Accept: text/markdown still goes through the edge HTML to markdown path, covering that too would need a middleware rewrite.

sdserranog (approved):
This is great! 🚢

app/api/toolkit-data/[toolkitId]/route.ts
```diff
@@ -30,6 +31,18 @@ export async function GET(
       );
     }
 
+    // Content-negotiate markdown: the toolkit page renders per-tool detail
+    // client-only, so the "copy as markdown" / agent view builds full markdown
+    // straight from the data here instead of from the slimmed HTML.
+    if ((request.headers.get("accept") ?? "").includes("text/markdown")) {
+      return new NextResponse(toToolkitMarkdown(data), {
```
Comment (sdserranog) [inline app/api/toolkit-data/[toolkitId]/route.ts:42]: nit: the same URL returns JSON or markdown depending on Accept, but the response is cached public with no Vary: Accept, so a CDN can serve the wrong representation to a later request.
Fix: add Vary: Accept to the response headers on this route.

Comment (juan-arcadedev) [reply app/api/toolkit-data/[toolkitId]/route.ts:42]: good catch — that would bite us behind a CDN. added Vary: Accept to the route's cache headers (covers both the JSON and markdown responses) in 67645300.

PR #1041, opened 2026-06-29
fix(docs): standardize account-email cookie name

juan-arcadedev:
## What
`PlaceholderReplacer` read the cookie `last_arcadedev_account_email`, but **nothing in the org sets that name**. Identity UI sets `last_arcade_account_email` (no "dev"), which is also what the marketing site (www) navbar reads. This switches docs to that canonical name.


sdserranog (approved):
:shipit:

PR #1065, opened 2026-07-16
Validate and canonicalize generated toolkit routes

jottakka:
## Summary
- validate every generated toolkit slug before it becomes a route
- make integration cards, static parameters, and the sitemap share that contract
- sort and deduplicate generated routes deterministically

## Test plan
- `pnpm test`

---

> [!NOTE]
> **Medium Risk**
> Touches integration routing, index link resolution, and sitemap/canonical SEO across many entry points; behavior is defensive and heavily tested but mis-slugging could hide or mis-link catalog entries.
> 
> **Overview**
> Unifies how integration **categories**, **slugs**, and **card URLs** are derived so the catalog, static routes, sidebar, sitemap, and canonical metadata stay in sync.
> 
> **Slug safety:** `getToolkitSlug` now rejects unsafe path segments and returns `null` when no valid slug exists; route listing, index resolution, canonical paths, and `llms.txt` skip those entries instead of emitting bad URLs.
> 
> **Shared contract:** `normalizeCategory` and the category allow-list move to a client-safe `toolkit-category` module; catalog enrichment prefers toolkit JSON over design-system metadata for `docsLink` and `category` (matching route helpers).
> 
> **Integrations index:** `resolveIndexToolkits` exposes a final `link` (including lone bare-name → `-api` remaps when no sibling owns the page), drops unsluggable catalog rows, and cards use that `link` rather than recomputing on the client.
> 
> **Data loading:** Toolkit JSON and `index.json` are validated with Zod; malformed index rows are filtered without dropping the whole index.
> 
> **SEO / discovery:** The sitemap merges authored MDX URLs with `listValidIntegrationLinks`, dedupes by URL, and omits `others` integration paths that redirect; toolkit pages only set `alternates.canonical` when a canonical path exists.
> 
> **Generator:** Sidebar sync reuses shared slug/category helpers and safer JSON escaping for generated `_meta.tsx` files.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bac2dceb9754740861ad14fc8efda4588091682e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/_lib/integration-index.ts
+++ b/app/_lib/integration-index.ts
@@ -13,6 +13,9 @@ export function toIntegrationLink(toolkit: {
   category?: string | null;
 }): string {
   const slug = getToolkitSlug({ id: toolkit.id, docsLink: toolkit.docsLink });
+  if (!slug) {
+    throw new Error(`Cannot build an integration link for: ${toolkit.id}`);
+  }
   const category = toolkit.category ?? "others";
   return `${INTEGRATIONS_BASE}/${category}/${slug}`;
 }
@@ -47,7 +50,12 @@ export function resolveIndexToolkits(
       continue;
     }
 
-    const link = toIntegrationLink(toolkit);
+    let link: string;
+    try {
+      link = toIntegrationLink(toolkit);
+    } catch {
+      continue;
+    }
     const hasPage = validLinks.has(link);
 
     // A bare duplicate of a real "-api" toolkit: drop it; the real card stays.
--- a/app/_lib/integration-index.ts
+++ b/app/_lib/integration-index.ts
@@ -23,7 +23,23 @@ export function toIntegrationLink(toolkit: {
   return `${INTEGRATIONS_BASE}/${category}/${slug}`;
 }
 
-export type ResolvedIndexToolkit = ToolkitWithDocsLink & { hasPage: boolean };
+export type ResolvedIndexToolkit = ToolkitWithDocsLink & {
+  hasPage: boolean;
+  /** Final card href (may remap bare names onto a real `-api` page). */
+  link: string;
+};
+
+const catalogOwnsLink = (
+  toolkits: ToolkitWithDocsLink[],
+  targetLink: string,
+  except: ToolkitWithDocsLink
+): boolean =>
+  toolkits.some((other) => {
+    if (other === except || other.isHidden) {
+      return false;
+    }
+    return toIntegrationLink(other) === targetLink;
+  });
 
 /**
  * Decide which catalog toolkits the integrations index should render, and
@@ -34,8 +50,10 @@ export type ResolvedIndexToolkit = ToolkitWithDocsLink & { hasPage: boolean };
  * no generated docs page — linking to them 404s. Given the set of links that
  * actually resolve (`validLinks`: dynamic toolkit routes + authored static
  * pages), this:
- *   - drops a bare entry when its `-api` sibling owns the real page (collapses
- *     Datadog/DatadogApi, Vercel/VercelApi, Ashby/AshbyApi, Customerio/...),
+ *   - drops a bare entry when another catalog toolkit owns the `-api` page
+ *     (collapses Datadog/DatadogApi, …),
+ *   - remaps a lone bare entry onto the `-api` page when that page exists but
+ *     no sibling catalog entry claims it,
  *   - de-dupes entries that resolve to the same URL (e.g. Notion/NotionToolkit),
  *   - flags the rest with `hasPage` so the caller can render doc-less toolkits
  *     as non-clickable cards instead of as broken links.
@@ -57,19 +75,28 @@ export function resolveIndexToolkits(
     if (!link) {
       continue;
     }
-    const hasPage = validLinks.has(link);
 
-    // A bare duplicate of a real "-api" toolkit: drop it; the real card stays.
+    let resolvedLink = link;
+    let hasPage = validLinks.has(link);
+
+    // Bare name with no page, but a real `-api` page exists.
     if (!hasPage && validLinks.has(`${link}-api`)) {
-      continue;
+      const apiLink = `${link}-api`;
+      if (catalogOwnsLink(toolkits, apiLink, toolkit)) {
+        // Sibling (e.g. DatadogApi) owns the page — drop the bare duplicate.
+        continue;
+      }
+      // Lone bare entry: surface the generated `-api` page instead of vanishing.
+      resolvedLink = apiLink;
+      hasPage = true;
     }
 
     // Collapse multiple catalog entries that point at the same URL.
-    if (seen.has(link)) {
+    if (seen.has(resolvedLink)) {
       continue;
     }
-    seen.add(link);
-    resolved.push({ ...toolkit, hasPage });
+    seen.add(resolvedLink);
+    resolved.push({ ...toolkit, hasPage, link: resolvedLink });
   }
 
   return resolved;
--- a/app/_lib/integration-index.ts
+++ b/app/_lib/integration-index.ts
@@ -1,3 +1,4 @@
+import { normalizeCategory } from "./toolkit-category";
 import { getToolkitSlug, type ToolkitWithDocsLink } from "./toolkit-slug";
 
 const INTEGRATIONS_BASE = "/en/resources/integrations";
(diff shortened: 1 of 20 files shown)


app/_lib/toolkit-data.ts
```diff
@@ -65,14 +69,41 @@ const DEFAULT_DATA_DIR = join(
...
+  );
+};
+
+const isToolkitIndexEnvelope = (
+  value: unknown
+): value is ToolkitIndexEnvelope =>
+  isRecord(value) &&
+  typeof value.generatedAt === "string" &&
+  typeof value.version === "string" &&
+  Array.isArray(value.toolkits);
```
kamath: cc @teallarson would these be better off as zod schemas?

teallarson: 100%

The generator already has a schema... we can't import that here because of module resolution BUT we can define one in app/_lib/toolkit-data.ts.

Something like:
```
// isToolkitIndexEntry becomes
const ToolkitIndexEntrySchema = z.object({
  id: z.string(),
  label: z.string(),
  version: z.string(),
  category: z.string(),
  type: z.string().optional(),
  toolCount: z.number().int().nonnegative(),
  authType: z.string(),
});

...
// isToolkitIndexEnvelope becomes
const ToolkitIndexEnvelopeSchema = z.object({
  generatedAt: z.string(),
  version: z.string(),
  toolkits: z.array(z.unknown()),
});

...
// isValidToolkitData becomes: 

const ToolkitDataSchema = z
  .object({
    id: z.string(),
    label: z.string().optional(),
    name: z.string().optional(),
    metadata: z.object({}).passthrough(),
  })
  .passthrough()
  .refine((v) => v.label !== undefined || v.name !== undefined);
  
  //etc.
```

cc: @jottakka !

jottakka:
Addressed the Zod / sitemap / shared-`normalizeCategory` review feedback in the latest commits.

Remaining card↔route category contract gaps (raw `toIntegrationLink` category, catalog not backfilling JSON category/`docsLink` precedence) are intentionally in the independent follow-up: https://github.com/ArcadeAI/docs/pull/1088

Re-requesting review for the Zod change on this PR.

app/_lib/integration-index.ts
teallarson, line 1: When a toolkit can’t produce a valid URL, this code throws an error in one place and catches it in another. That works, but it’s harder to follow than needed. Simpler approach: if we can’t build a URL, just skip that toolkit and move on — no throw/catch needed.

app/_lib/toolkit-static-params.ts
```diff
@@ -70,6 +77,9 @@ export function getToolkitCanonicalPath(toolkit: {
 }): string {
   const category = normalizeCategory(toolkit.category);
   const slug = getToolkitSlug({ id: toolkit.id, docsLink: toolkit.docsLink });
+  if (!slug) {
+    throw new Error(`Cannot build a canonical path for toolkit: ${toolkit.id}`);
+  }
```
teallarson, line 60: similar to the comment above: invalid slugs throw errors that get caught elsewhere. Would be cleaner if “bad slug” just meant “don’t include this page” everywhere, consistently.

app/_lib/integration-index.ts
teallarson: Also: Sidebar and canonical URLs now sanitize categories (so bad values like ../../outside get mapped to others), but integration card links don’t seem to do that yet. Should those use the same category cleanup so cards and pages always match?

app/_lib/toolkit-data.ts
teallarson, line 1: This adds a fair amount of validation code. Is that solving a real problem we’ve hit (bad rows in index.json), or is it defensive? 

If we haven’t seen bad data in practice, we might get most of the benefit from just skipping toolkits with invalid slugs.

tests/sitemap-toolkit-routes.test.ts
teallarson, line 1: tests/sitemap-toolkit-routes.test.ts and the new checks in tests/sitemap.test.ts seem to test the same thing (toolkit URLs in sitemap, excluding others). Could we keep one of these and delete the other?

toolkit-docs-generator/tests/scripts/sync-toolkit-sidebar.test.ts
teallarson, line 1: The new sidebar tests for bad slugs / bad categories overlap a lot with tests already added in toolkit-slug.test.ts and toolkit-static-params.test.ts. 

I’d keep the sidebar tests focused on “does the generated sidebar file look right?” and drop the ones that repeat slug logic tests.

app/sitemap.ts
```diff
@@ -43,11 +44,27 @@ async function collectRoutes(dir: string): Promise<MetadataRoute.Sitemap> {
...
 
+async function collectToolkitRoutes(): Promise<MetadataRoute.Sitemap> {
+  const routes = await listToolkitRoutes();
+  return routes
+    .filter(({ category }) => category !== "others")
+    .map(({ category, toolkitId }) => ({
+      url: `${NORMALIZED_SITE_URL}/en/resources/integrations/${category}/${toolkitId}`,
+    }));
+}
+
```
teallarson, line 63: We already have a helper that builds the list of valid integration page URLs (listValidIntegrationLinks). The sitemap is rebuilding those URLs manually. Can we reuse the existing helper instead?

teallarson (changes requested):
This PR is doing the right thing overall. Making sure toolkit page URLs are built the same way everywhere (cards, sitemap, sidebar, etc.). 

The main feedback is: let's please clean up some duplicated logic and tests in here 🙏

app/_lib/integration-index.ts
jottakka: Done — `toIntegrationLink` now returns `null` for bad slugs, and `resolveIndexToolkits` just skips those entries (no throw/catch).

jottakka: Agreed — cards now use the same client-safe `normalizeCategory` helper as routes/sidebar (`app/_lib/toolkit-category.ts`), so unknown/empty categories map to `others` consistently.

app/_lib/toolkit-static-params.ts
jottakka: Done — `getToolkitCanonicalPath` now returns `null` instead of throwing; callers skip / omit canonical when there is no safe slug.

app/_lib/toolkit-data.ts
jottakka: Mostly defensive + consistency with the earlier request to use Zod here (generator already validates with Zod; we can’t import those schemas into `app/_lib`).

Practical benefit we kept: envelope validates, then **filters** bad index rows instead of dropping the whole catalog. Happy to slim this further to “skip invalid slugs only” if you’d rather — the slug skip path is now consistent everywhere either way.

tests/sitemap-toolkit-routes.test.ts
jottakka: Kept `tests/sitemap-toolkit-routes.test.ts` (mocked, focused on others exclusion + helper reuse) and removed the overlapping toolkit-URL loop from `tests/sitemap.test.ts`.

toolkit-docs-generator/tests/scripts/sync-toolkit-sidebar.test.ts
jottakka: Removed the overlapping bad-slug / unknown-category sidebar cases — those live in `toolkit-slug` / `toolkit-static-params` / `toolkit-category` tests. Sidebar tests now stay focused on generated `_meta` shape + sync behavior (kept the “href categories match static params” integration check).

app/sitemap.ts
jottakka: Done — sitemap now builds toolkit URLs from `listValidIntegrationLinks()`, filters out `others` (redirects), and falls back to authored pages only if toolkit listing throws.

app/_lib/toolkit-data.ts
jottakka: Implemented the app-local Zod schemas as sketched (envelope + per-entry filter). See also the later note about how defensive this is — open to simplifying if you’d prefer slug-skip only.

jottakka:
Addressed Teal’s latest review comments in the newest commits:

- no throw/catch for bad URLs — return `null` and skip
- cards use shared `normalizeCategory` (folded #1088 into this PR)
- sitemap reuses `listValidIntegrationLinks` (excludes `others`)
- deduped sitemap + trimmed overlapping sidebar slug tests
- left a note on the Zod thread about defensive vs slug-skip-only

#1088 can be closed as superseded.

PR #1093, opened 2026-07-24
feat: capture signup attribution in docs middleware

juan-arcadedev:
Captures `utm_*`/`gclid`/referrer on the first docs page a visitor lands on and writes the shared `.arcade.dev` cookie that identity-ui reads back at registration, giving docs the same signup attribution as identity-ui and www. Uses `@arcadeai/ui-kit/utils/attribution` (v0.14.0 — 0.15.0 needs design-system ≥7.1.0, which isn't published yet) so the cookie format stays in parity across repos. The cookie attaches to every `proxy.ts` response, including the locale redirects, since those drop the query string on first visit. Adds a test covering param capture and the cookie contract identity-ui parses.

Note: pulling in ui-kit dragged ~29 transitive packages into the lockfile via pnpm auto-install-peers (Vercel AI SDK, react-query, react-router, arcadejs); none are imported by the attribution module or docs — a standalone attribution lib would avoid this.

---

> [!NOTE]
> **Medium Risk**
> Cross-subdomain cookie behavior affects marketing attribution and must stay aligned with identity-ui; lockfile bloat adds supply-chain surface though runtime usage is limited to attribution utils.
> 
> **Overview**
> Docs middleware now records **first-touch signup attribution** (`utm_*`, `gclid`, referrer) into the same cookie identity-ui uses at registration, so campaigns that land on docs are not lost when users sign up elsewhere on `arcade.dev`.
> 
> A new `withAttribution` wrapper runs on **every** `proxy` response (including locale and `/toolkits` redirects) so attribution is persisted before redirects strip the query string. Cookie scope is `.arcade.dev` in production and host-only on preview/local; an existing cookie is left unchanged.
> 
> Shared helpers come from **`@arcadeai/ui-kit@0.14.0`** (`extractAttribution` / `buildAttributionCookie`). The lockfile grows with ui-kit’s peer-related transitive deps; docs only imports the attribution utils.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d16d40fa4db9c1ab260d90d26706d2b1fa1c2ddb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tests/attribution.test.ts
```diff
@@ -0,0 +1,61 @@
+import {
```
sdserranog: I think we already tested most of this in ui-kit, so there's no need to duplicate it here.

juan-arcadedev: True, removed.

proxy.ts
```diff
@@ -1,7 +1,55 @@
...
+    ? SHARED_COOKIE_DOMAIN
+    : undefined;
+}
+
+// Capture signup attribution (utm_*/gclid/referrer) on the first docs page a
+// visitor lands on with those params, into a first-party cookie the identity-ui
+// register flow reads back. First-touch wins: skip if the cookie already exists.
+// Attaches to whatever response we return because most first visits hit the
+// locale redirect below, which drops the query string.
+function withAttribution(
```
sdserranog, line 30: nit: if we ended up adding this a third time, I think we should add this fn to ui-kit. I don't think it's worth doing that now because we only have it in identity-ui and here.

sdserranog (approved):

proxy.ts
juan-arcadedev: Agreed, I'll just go ahead and add it when workign on the dashboard attribution aswell and then will do a follow up and remove it from here and on the landing page

PR #1101, opened 2026-07-29
fix(integrations): correct Nimble page tool list and naming

thierrypdamiba:
Follow-up to #1097 — the "improved version" from the Slack thread, since #1097 was intentionally merged as-is to unblock shipping the URL.

**Verified against the live server**, not just published docs: `tools/list` handshake against `mcp.nimbleway.com/mcp` (Nimble 3.4.3, protocol 2025-06-18) using our trial account.

### The tool list was wrong: 27 tools, not 6

Nimble's docs describe six capability families, and the page reproduced those as six rows. The live server exposes **27 tools** — each family ships multiple tools for synchronous, asynchronous, and status/listing calls.

This matters in Arcade specifically. The gateway tool picker lists *individual* tools, so someone who reads "6 capabilities" and then sees 27 checkboxes has no way to map between them. The table now carries a per-capability tool count and the total:

| Capability | Tools |
| --- | --- |
| Search | 1 |
| Extract | 2 |
| Crawl | 4 |
| Map | 1 |
| Extract templates | 7 |
| Web search agents | 11 |
| Task results | 1 |

Also added **web search agents** as a capability, which the original page omitted entirely — it's the largest family on the server at 11 tools.

### Async polling gotcha

Many Nimble tools return a task ID and require polling a companion tool (`nimble_task_results`, `nimble_agents_run_status`, `nimble_crawl_status`). Since selecting a subset of tools is the normal gateway workflow, it's easy to build a gateway with only the starting tools and leave your agent kicking off jobs it can't read back. Now called out.

### Tool naming was misleading

The page said tools appear as `<your-server-id>.<tool>`. Nimble prefixes its tools with `nimble_`, so with server ID `nimble` the real name is `nimble.nimble_search` — a doubled prefix that surprises people mid-integration. Confirmed `nimble_search` exists on the live server.

### Cost section: attributed, not asserted

I initially removed this believing it was unsourced. **That was my error** — I'd only checked `docs.nimbleway.com`. It's published on nimbleway.com/mcp and in Nimble's MCP blog post:

> "Compound accuracy over time through a Proprietary Index and Memory that gets smarter with every query"
> "Cut token costs by retrieving exactly what's needed — no redundant searches, no parsing raw pages with an LLM"

Restored with two adjustments:

1. **Attributed to Nimble** rather than asserted as fact — product claims about a third party's internals that we can't verify from outside, on a partner-facing page.
2. **Dropped the detail that outran the source.** The original said the cache stores "sources, search history, and responses" and the memory layer "extracts key learnings after each run and feeds them into the next search plan." Nimble publishes "Proprietary Index and Memory" and "auditable Search Plans" — not that breakdown.

Kept the practical MCP token caveat alongside it. The two sources genuinely conflict: the product page sells retrieval efficiency, while Nimble's docs warn MCP "can consume tokens quickly" and point high-volume users at their CLI skill. Readers should see both.

### Stale tool names on Nimble's marketing page

Worth passing back to Nimble: nimbleway.com/mcp advertises `nimble_deep_web_search`, `nimble_google_maps_search`, and `nimble_google_maps_reviews`. **None of these exist on the live server.** Their docs were right and their product page is out of date — a heads-up they'd probably want, since it's the page prospects read.

### Style

Changelog said "Partner MCP Server"; STYLEGUIDE.md specifies lowercase "MCP server". Page body was already correct.

### Verified, not changed

- `Authorization: Bearer ${secret:NIMBLE_API_KEY}` matches both Nimble's "store the raw key, Bearer prefix goes in the header" instruction and the `${secret:NAME}` syntax in add-remote-servers — confirmed working against the live server
- All external links and internal anchors resolve, including `#configure-advanced-settings`
- `pnpm build` passes, route compiles, Vale clean (two remaining warnings are pre-existing false positives on UI labels after colons)
- Nimble absent from `search/_meta.tsx`, but so is Tavily — partner pages are reached via the catalog card, existing pattern
- `llms.txt` gives Nimble the placeholder "Documentation page", but that hits 21 entries including Tavily (generator fallback when summarization fails). Systemic, worth its own fix.

---

> [!NOTE]
> **Low Risk**
> Documentation-only changes to the Nimble integration page and changelog; no runtime, auth, or API behavior changes.
> 
> **Overview**
> Updates the **Nimble** partner integration doc and a one-line **changelog** style fix so they match the live MCP server and Arcade gateway behavior.
> 
> The capabilities section now reflects **seven families** (including **web search agents**), explains that the tool picker shows **~two dozen** individual tools—not one row per capability—and documents **`nimble_` prefixes** and names like `nimble.nimble_search`. A callout warns gateway builders to include **async polling** tools (`nimble_task_results`, `nimble_agents_run_status`) with starters, and the page notes which tools **write** to the Nimble account.
> 
> **Cost and token usage** replaces the old caching narrative with Nimble-attributed retrieval claims, practical MCP token guidance (narrow tool selection, prefer search/extract), and a pointer to Nimble’s CLI skill for high-volume use.
> 
> Changelog entry wording changes **Partner MCP Server** to **Partner MCP server** for style consistency.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b1d6e47ef5ad844c813123a501ab8905571abc87. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/en/references/changelog/page.mdx
+++ b/app/en/references/changelog/page.mdx
@@ -13,7 +13,7 @@ _Here's what's new at Arcade.dev!_
 
 **Misc**
 
-- `[feature - 🚀]` Add Nimble as a Partner MCP Server for real-time web data in the search integrations catalog.
+- `[feature - 🚀]` Add Nimble as a Partner MCP server for real-time web data in the search integrations catalog.
 
 ## 2026-07-24
 
--- a/app/en/resources/integrations/search/nimble/page.mdx
+++ b/app/en/resources/integrations/search/nimble/page.mdx
@@ -32,17 +32,19 @@ Set these under **Advanced settings → Custom headers** when you add the server
 
 ## What you can do
 
-Once you register Nimble in your Arcade project, Arcade discovers its tools automatically. Nimble exposes 27 tools, because most capabilities ship more than one tool: a synchronous call, an asynchronous variant, and status or listing calls. They group into these capabilities:
+Once you register Nimble in your Arcade project, Arcade discovers its tools automatically. Nimble groups them into these capabilities:
 
-| Capability | Tools | What it does |
-| --- | --- | --- |
-| Search | 1 | Real-time web search with configurable content richness. |
-| Extract | 2 | Pull structured content from a specific URL, synchronously or asynchronously. |
-| Crawl | 4 | Crawl multiple pages on a site, then check progress, list jobs, or cancel a run. |
-| Map | 1 | Discover the URLs on a site by crawling its pages and sitemap. |
-| Extract templates | 7 | Browse the template catalog, run templates against a URL, and generate or refine custom ones. |
-| Web search agents | 11 | Create, update, and run research agents, then poll runs and read their results. |
-| Task results | 1 | Fetch the status and results of any async task. |
+| Capability | What it does |
+| --- | --- |
+| Search | Real-time web search with configurable content richness. |
+| Extract | Pull structured content from a specific URL, synchronously or asynchronously. |
+| Crawl | Crawl multiple pages on a site, then check progress, list jobs, or cancel a run. |
+| Map | Discover the URLs on a site by crawling its pages and sitemap. |
+| Extract templates | Browse the template catalog, run templates against a URL, and generate or refine custom ones. |
+| Web search agents | Create, update, and run research agents, then poll runs and read their results. |
+| Task results | Fetch the status and results of any async task. |
+
+Most capabilities ship more than one tool: a synchronous call, an asynchronous variant, and status or listing calls. Expect roughly two dozen individual tools in the tool picker rather than seven. The exact set depends on your Nimble plan and on the account you connect with, so [connect as an admin user](/guides/mcp-gateways/add-remote-servers#save-and-confirm-the-connection) to discover the broadest list, and check the picker for what your project actually has.
 
 Nimble prefixes its own tools with `nimble_`, and Arcade namespaces each tool under the server ID you choose when you add the server. With the server ID `nimble`, tool names appear as `nimble.nimble_search` in the Playground, the gateway tool picker, and SDK calls. Copy the exact name from the tool picker before calling it.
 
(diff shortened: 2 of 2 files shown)


ilchemla:
Thanks for taking this on, @thierrypdamiba, and sorry for the slow reply from our side. Read it end to end. The structural additions are right and better than what was there: the async/polling callout is the most useful thing on the page, the read-vs-write split is accurate, and "roughly two dozen individual tools" matches what the server exposes today (27).

Four notes.

1. The Search row lost `focus`, which is the parameter that most differentiates Nimble from generic search. "Configurable content richness" describes `search_depth` accurately (close to our own spec wording for it), but `focus` is separate and worth keeping: `general`, `news`, `shopping`, `social`, `geo`, `location`, or an explicit list of subagent names. Suggested: "Live web search with configurable focus (general, news, shopping, social, geo, location) and content richness."

2. Product naming: we write **Web Search Agents** capitalised as a product name, and the `nimble_extract_templates_*` tools as **Templates**. Small thing, but those are the names in our docs and pricing.

3. On tool selection, the guidance we would give is choose by job, not by breadth. Search + Extract are the fast raw web data layer when your agent owns the reasoning. Web Search Agents are the managed research layer when you want a completed task back: they run the search, browsing, extraction, reasoning, citations, and schema enforcement in one call. "Prefer the narrowest tool" reads as though a Web Search Agent is a heavier version of search, when it is a different job. Suggested: "Pick by job. Search or Extract when your agent reasons over raw results; a Web Search Agent when you want a synthesized, cited, schema-constrained answer."

4. On the cost section: the mechanism behind the efficiency claim, if it is useful to state plainly, is that memory records which retrieval paths found the right data and a private index stores prior outputs, sources, and search history, so Nimble does not rediscover where information lives on every run. That is the substance behind "gets smarter with every query" if you would rather describe it than quote it.

One question: "the exact set depends on your Nimble plan and on the account you connect with" — could you point me at where that came from? I want to make sure our own docs say the same thing.

Otherwise this reads accurately to us. `nimble_task_results` and `nimble_agents_run_status` are both correct as written and the CLI-skill link resolves. Once those are settled it is good to merge from our side.

PR #1104, opened 2026-07-31
fix: restore static rendering and index toolkit pages (GRO-274)

teallarson:
This PR restores static rendering for the docs site and adds generated toolkit pages to `sitemap.xml`.

### Static rendering

The root layout previously awaited `headers()` to derive a locale that is always `en`. Because `headers()` makes the root route tree dynamic, this prevented the docs pages—including generated toolkit pages—from being prerendered.

The layout now uses the constant locale directly. This is safe while `proxy.ts` redirects every supported non-English or unlocalized path to `/en`. Restoring real i18n should use an `app/[lang]/` route segment rather than request-header parsing.

### Sitemap coverage

The sitemap directory walk skips `[toolkitId]` directories because they are dynamic route segments, so generated toolkit pages were missing from the sitemap. The sitemap now reuses `listValidIntegrationLinks()`, the same route enumeration used by the integrations index, and deduplicates authored static partner pages.

Generated toolkit entries use the newest toolkit JSON modification time, so changes to committed toolkit content advance their sitemap timestamps. Hidden toolkits remain excluded.

Sitemap generation is best effort: if toolkit enumeration fails, static sitemap routes are still returned and a warning is emitted. Normal page/static-param validation continues to fail loudly for invalid toolkit data.

### Verification

- `pnpm build` passes and generated toolkit pages remain prerendered.
- Targeted sitemap and proxy tests pass (10 tests).
- Tests cover representative toolkit categories, hidden-toolkit exclusion, timestamp tracking, redirect collisions, sitemap fallback behavior, and locale redirects.

### DIFF
--- a/app/layout.tsx
+++ b/app/layout.tsx
@@ -10,7 +10,6 @@ import { TranslationBanner } from "@/app/_components/translation-banner";
 import "@/app/globals.css";
 import { Discord, Github } from "@arcadeai/design-system";
 import { GoogleTagManager } from "@next/third-parties/google";
-import { headers } from "next/headers";
 import Link from "next/link";
 import Script from "next/script";
 import { Head } from "nextra/components";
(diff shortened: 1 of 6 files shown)


teallarson:
@cursor review

app/layout.tsx
teallarson, line 1: I'm opting to remove these things until we move forward with i18n more holistically. 

Right now it's just adding complexity with no plan (that I've heard of) in the immediate future to leverage it.

sdserranog (approved):
:shipit:

PR #1105, opened 2026-07-31
refactor: singular + enforced toolkit data contract

teallarson:
This PR makes redirects and toolkit data use shared, typed sources instead of duplicated definitions and text parsing.

## Before → after

| Area | Before | After |
| --- | --- | --- |
| Redirects | Data embedded in a 962-line `next.config.ts`; consumers parsed the file as text. | Redirects live in `redirects.ts`; consumers import the data directly. The 156-entry array is unchanged. |
| Redirect checking | The tested helper lived in an unused copy. | The shipping checker imports the tested helper module. |
| Toolkit primitives | Data-directory paths, ID normalization, categories, and slug logic were repeated across the app and generator. | Shared primitives have one definition and all consumers use `TOOLKIT_DATA_DIR` consistently. |
| Data contract | The app used a small duck-check and then cast JSON to `ToolkitData`. | The app and generator share Zod schemas; types are inferred from the same contract. |
| Invalid data | Missing files, invalid JSON, and invalid shapes were handled the same way and could silently hide a toolkit. | Missing data returns `null`; invalid JSON or schema data throws with the file path and validation error. |
| Category safety | Unknown categories fell back to an unreachable `others` route. | Missing categories produce non-clickable entries; unknown categories fail clearly; `others` is removed. |
| Toolkit reads | Route generation repeatedly scanned the toolkit directory. | One cached loader indexes the directory once per process. |

## Compatibility

- Redirect output is unchanged: 156 entries, same content and order.
- Generated pages and routes remain unchanged.
- The cache removes redundant directory passes; it is not presented as a measured build-time speedup.

## Verification

- `pnpm test` passes.
- `pnpm exec tsc --noEmit` passes.
- `pnpm lint` reports no errors and the same three pre-existing complexity warnings.

One intentional trade-off: a corrupt toolkit file now fails the directory load instead of affecting only that toolkit. In production, the build catches this before requests are served.

---

> [!NOTE]
> **Medium Risk**
> Changes affect static route generation, sitemap/redirect correctness, and production toolkit API reads—invalid JSON now fails builds/requests instead of hiding toolkits, which is safer but stricter.
> 
> **Overview**
> **Redirects** move out of `next.config.ts` into importable `redirects.ts` (same rules, no behavior change). CI, pre-commit, `check-redirects`, `update-links`, and sitemap tests now use that module instead of regex-parsing config.
> 
> **Toolkit docs data** is a single contract: UI types are `z.infer` from generator `toolkit-schemas`; `readToolkitFile` / `readToolkitIndex` validate with Zod. Missing files still return `null`; bad JSON or schema mismatches **throw** with paths so builds break on corrupt nightly output. `loadAllToolkitData` + production lookup caching dedupe reads; slug vs normalized id cache keys stay distinct for the API route.
> 
> **Routing / integrations**: slug, category, and data-dir helpers live in `toolkit-docs-generator/src/shared/`; the fake **`others`** catch-all is gone—no category means `toIntegrationLink` returns `null` and cards are non-clickable; unknown categories throw. Sidebar sync and filters follow the same category list (design-system `CATEGORIES` for filters).
> 
> **DX / ops**: webpack `extensionAlias` for shared `.js` imports; `zod` is a runtime dependency; toolkit docs CI gets a Slack alert job on generation failure; new tests cover data dir, cache, parity, and category route dirs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 53e7df1515675eb4f261b0934addb92b12bf210c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


teallarson:
@cursor review

sdserranog (approved):
🚢

PR #1111, opened 2026-08-03
docs(generator): document data provenance and source-of-truth rules

teallarson:
> **On hold — do not merge. Superseded by #1112.**

## Purpose

This draft records the provenance and source-of-truth rules for the toolkit documentation pipeline. It explains how upstream API data, design-system metadata, generated enrichment, and hand-authored curation become the JSON consumed by the docs app.

## How it fits

```text
API + design metadata + LLM enrichment + curation
                         │
                         v
                      merger
                         │
                         v
                 toolkit JSON -> docs app
```

The document is intended to make field ownership and durability clear when debugging generated output or changing a pipeline stage. It covers the shared schema, category and URL derivation, toolkit grouping, corruption handling, exclusions, caching, and metadata completeness.

## Status

The original draft described hand-authored prose as living only in generated JSON and warned that forced regeneration could delete it. #1112 now gives that prose a durable `curation/` source, so this PR's prose-hazard section is outdated.

This should be rewritten as the narrower, post-#1112 provenance document before it is considered for merge. The current draft is retained only as a reference for that rewrite.


teallarson:
outdated

PR #1115, opened 2026-08-06
docs(snowflake): add Snowflake auth provider page

avoguru:
Scoped to the auth provider docs only, now that the auto-generated toolkit docs have merged.

Snowflake is set up as a custom OAuth 2.0 provider, because its OAuth endpoints are specific to each account. This page covers that: creating the security integration in Snowflake, registering it in Arcade, setting the `SNOWFLAKE_ACCOUNT_SUBDOMAIN` secret, and giving each person a role.

Each end user signs in to Snowflake themselves, so every query runs under that user's own role.

Also adds the provider to the auth-providers catalog page.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Low Risk**
> Documentation and integrations index display logic only; no auth runtime or API behavior changes.
> 
> **Overview**
> Adds a **Snowflake** auth provider reference page and lists it on the auth-providers catalog with a card linking to `/references/auth-providers/snowflake`.
> 
> The new guide walks through per-user OAuth via a Snowflake security integration (provider ID `snowflake`), Arcade custom OAuth endpoints (`/oauth/token-request`), the `SNOWFLAKE_ACCOUNT_SUBDOMAIN` secret, role grants, MCP gateway usage, and a sample `OAuth2(id="snowflake")` tool.
> 
> On the integrations index, **`resolveIndexToolkits`** now sets **`isComingSoon: !hasPage`** so shipped toolkits with docs are not stuck behind a stale design-system flag; **`toolkits-client`** passes **`isComingSoon`** from that resolved value only.
> 
> **`public/llms.txt`** is regenerated to include the Snowflake auth provider link.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 52256e0aa712a3670ea1f4c2fbeb7ecbd64158ae. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #1116, opened 2026-08-06
fix(integrations): derive "coming soon" from whether a page exists

avoguru:
The Snowflake tile still reads "Coming soon" even though its docs generated and merged in #1114.

The index took `isComingSoon` from the design-system catalog, which is a hand-maintained constant in a published package. It goes stale the moment a toolkit ships, and the generator copies it straight into the toolkit JSON, so a nightly run keeps writing it back.

This derives the flag from `hasPage` instead. The badge means "there is nothing to read yet", which is what `hasPage` already answers, and it no longer waits on a package release. **One tile changes: Snowflake.**

It also makes the badge and the sort agree. `use-toolkit-filters` sorted on the raw design-system flag while the card rendered on `isComingSoon || !hasPage`, so seven toolkits (CloseIO, Discord, GoogleForms, Notion, Obsidian, Twilio, Twitch) showed a "Coming soon" badge while sorting among the available ones. They now sort with the other unavailable toolkits.

Verified locally: the Snowflake card renders "Arcade Optimized MCP Server" with no badge and links through to the toolkit page. Every entry now satisfies `isComingSoon === !hasPage`. 796 tests pass.

The design-system entry is still stale (`isComingSoon: true`, `category: "development"`) and worth correcting separately, but the docs site no longer depends on it for this.

---

> [!NOTE]
> **Low Risk**
> Small display/sort fix on the integrations index; no auth, data, or API changes.
> 
> **Overview**
> **Coming soon** on the integrations index is now **`isComingSoon: !hasPage`** in `resolveIndexToolkits`, instead of the design-system catalog flag that can stay `true` after docs ship (e.g. Snowflake).
> 
> `ToolkitsClient` passes **`toolkit.isComingSoon` only**—it no longer ORs with `hasPage === false`, because resolution already aligns badge, click behavior, and **`use-toolkit-filters`** sort on the same value.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3903d32b1282bb1f94576be9cac197d8a0027396. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/_lib/integration-index.ts
+++ b/app/_lib/integration-index.ts
@@ -66,7 +66,15 @@ export function resolveIndexToolkits(
       }
       seen.add(link);
     }
-    resolved.push({ ...toolkit, hasPage });
+
+    // "Coming soon" means there is nothing to read yet, so derive it from
+    // whether a page exists rather than trusting the design-system flag. That
+    // flag is a hand-maintained constant in a published package, so it goes
+    // stale the moment a toolkit ships and keeps a live toolkit looking
+    // unavailable. Deriving it here also keeps the badge and the sort order
+    // (`use-toolkit-filters`) reading the same value, which they previously
+    // did not.
+    resolved.push({ ...toolkit, hasPage, isComingSoon: !hasPage });
   }
 
   return resolved;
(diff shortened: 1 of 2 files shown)


avoguru:
Folded into #1115 — the auth provider page and this fix belong together.

PR #1125, opened 2026-08-12
docs: document the toolkit curation format and add an authoring skill

teallarson:
Closes GRO-278. Follow-up from @sdserranog's review of #1112: `curation/` had no format reference, only examples.

## What's here

- **`toolkit-docs-generator/CURATION.md`** — the reference. File kinds, every frontmatter key, slot-order tables, the authoritative-directory rule, and each failure message with its trigger.
- **`.claude/skills/curate-toolkit-docs/`** — the procedure for adding or editing curation.
- **`validate-curation` CLI** — compiles a curation tree with the same code generation uses, so authors can check their work without Engine credentials. It can't verify `tool:` targets, which need the live tool list; the command says so.

## Documented, not changed

Writing the tables surfaced places where behavior contradicts naming. All left as-is, all now written down:

- `position` is a slot name, not a spatial relation — `description` + `after` renders *above* the summary.
- Some accepted combinations render nowhere (`replace` on several locations; `parameters`/`secrets`/`output` at toolkit level).
- Filenames don't order anything; the renderer sorts by `priority`. Every existing file has a `001-` prefix, so this misleads.
- `type: section` renders as a callout — all 26 existing ones sit in boxes.
- `imports/` and `pages/` reach the JSON but nothing in the app reads them.

Happy to file tickets if we'd rather fix any of these.

## Note

`compileCurationDirectory` is extracted from `MarkdownCurationSource.loadData` so validation can report every broken toolkit instead of dying on the first. `loadData` still throws on the first error, so generation is unchanged.

## Verification

Typecheck, lint, and 834 tests pass. `validate-curation` is clean across all 75 toolkit directories, and each failure path was exercised by hand. Docs and tooling only — no rendered-page changes.

### DIFF
--- a/.claude/skills/curate-toolkit-docs/SKILL.md
+++ b/.claude/skills/curate-toolkit-docs/SKILL.md
@@ -1,0 +1,136 @@
+---
+name: curate-toolkit-docs
+description: Add or edit hand-authored prose on an Arcade toolkit reference page by writing curation files under toolkit-docs-generator/curation/. Use when asked to document a toolkit's auth setup, explain an enum or parameter, add a warning to a tool, or fix wording on a toolkit page — anything where the text belongs to a specific toolkit or tool rather than a standalone docs page.
+---
+
+# Curating toolkit docs
+
+Toolkit reference pages are generated. Editing the JSON under
+`toolkit-docs-generator/data/toolkits/` does nothing durable — the next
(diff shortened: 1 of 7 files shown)


teallarson:
@bugbot review

sdserranog (approved):
Amazing! 🚢

PR #1129, opened 2026-08-15
docs: document tool executions in the Operate area

evantahler:
Adds a **Tool Executions** page under Operate → Governance, next to Audit Logs, documenting the dashboard surface and read API shipped in [ArcadeAI/monorepo#2661](https://github.com/ArcadeAI/monorepo/pull/2661) and [#2662](https://github.com/ArcadeAI/monorepo/pull/2662).

The page covers both jobs the feature exists for: a developer debugging why a tool call failed by reading the exact inputs and outputs, and an operator reviewing what a project ran for compliance and usage. It documents the access split those PRs introduced — the history is readable by anyone holding a role on the project, while the recorded inputs and outputs take project-admin authority, and diagnostics stay readable either way — plus the dashboard filters, the three read endpoints, and the retention window.

Also adds three dashboard screenshots, cross-links the page from the developer-facing tool error handling guide, and adds it to the Governance and Operate card grids.

Verified with `pnpm lint`, `pnpm test` (835 pass), `pnpm check-meta`, Vale (0 errors), and a local dev-server render of the new page and both card grids.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Low Risk**
> Documentation-only changes with no runtime, auth, or application code impact.
> 
> **Overview**
> Adds **Tool Executions** under Operate → Governance, documenting the dashboard and read API for per-project tool run history (debugging failures and compliance/usage review).
> 
> The new page explains what gets recorded, dashboard filters and detail views (including admin-only **Show inputs/outputs** toggles), the member vs project-admin access split, list/count/detail API endpoints with `include_inputs`/`include_outputs` behavior, and org-wide recording and retention via **Logging Policy**.
> 
> Governance and Operate hub copy now mentions reviewing tool runs alongside audit logs; nav `_meta` and card grids link to the page. The build **Tool error handling** guide gains a short section pointing operators to Tool executions for post-mortems. **`public/llms.txt`** lists the new URL for LLM indexing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 691518224ac285fd329644ea6ea96eb320c8bd79. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #1141, opened 2026-08-25
docs(tool-executions): recording and retention are self-service [PLT-2978]

josebarrueta:
Stacked onto #1129 so nothing on that branch is rewritten — merge this into it and the page is current.

## Why

The **Retention** section describes the world before [PLT-2978](https://linear.app/arcadedev/issue/PLT-2978). It says:

> To change the retention window, or to turn off execution logging and delete the history for your organization, contact Arcade support.

That capability is being shipped as self-service in a stack now in review (`monorepo` #3165 the API, #3243 the dashboard page). The stale instruction is the obvious problem, but three smaller ones matter more, because each would mislead a reader who acts on it.

**It offered project-level overrides** — "7 days unless your organization sets a different default or a project overrides it". Project-level settings are an explicit non-goal of the spec. Nothing customer-reachable sets a project's own window, and the API reports `retention_changeable_at_project: false` for exactly this reason. A reader would go looking for a control that does not exist.

**It folded two different actions together** — "turn off execution logging and delete the history". They are not the same operation and behave differently. Turning recording off stops new runs being written down and leaves what is stored to expire on the window; the retention sweep marks on the window alone and never reads whether recording is on. Shortening the window is the control that deletes, and it reaches records already held. Someone who turns recording off expecting an erasure gets neither the deletion nor a warning.

**It gave no bounds**, so nothing told a reader that `0` and `400` are refused with a `422` rather than clamped.

## What changed

One file, `app/en/operate/governance/tool-executions/page.mdx`. The section is now **Recording and retention** and covers:

- the two settings, both organization-wide, and that an organization admin sets them
- how to change them — the dashboard page and `PUT /v1/orgs/{org_id}/logging-config`
- the bounds, and that out-of-range is a `422`
- **what each setting does to history you already have** — the distinction above, stated plainly, including that shortening the window cannot be undone
- what a reader sees when the list is empty, since an empty history has several causes and the product now says which applies

The intro also picks up a pointer, since "Arcade records every tool run" is only true while recording is on.

## Checked

`bun run build` passes. It caught a real mistake first: I used `<Note>`, which this site does not have — the convention is `<Callout>` (282 uses) and it needs an explicit `import { Callout } from "nextra/components"`. Prerendering failed until both were right.

Every claim is checked against the implementation rather than the spec prose:

| Claim | Source |
| -- | -- |
| Default 7 days, max 90 | `ORG_DEFAULT_LOG_RETENTION_DAYS`, `MAX_LOG_RETENTION_DAYS` in `core/conf.py` |
| Out-of-range is 422 | `within_platform_maximum` on the FastAPI-validated body |
| Recording off does not delete | `runMarkPhase` reads `LogRetentionDays` only, never `LoggingEnabled` |
| Shortening deletes retroactively | the mark phase re-evaluates all unmarked rows against the current window |
| Org-admin authority | `logging_config_update`, org-admin only, plus `can_edit_logging` on the org read |

## Sequencing

This documents behaviour that is **not merged yet**. #3165 and #3243 are green and awaiting review, not released. If #1129 is likely to ship first, hold this one until the stack lands — otherwise the page promises a dashboard page that isn't there. Happy to gate it however you prefer.

---

> [!NOTE]
> **Low Risk**
> Documentation-only change with no runtime or API behavior modified in this PR.
> 
> **Overview**
> Updates the **Tool Executions** governance doc so it matches org-wide, self-service execution logging instead of “contact support” and misleading retention behavior.
> 
> The intro now states that recording is on by default and points to the new **Recording and retention** section. That section replaces the old **Retention** block: it describes two org-admin settings (recording on/off and retention days), Cloud defaults (7-day window, 90-day max) vs self-hosted, and how to change them via the dashboard **Execution Logging** page or `PUT /v1/orgs/{org_id}/logging-config`. It documents validation (`422` for out-of-range values, minimum 1 day, zero is not valid) and clarifies that **turning recording off** only stops new writes while existing rows expire on the window, whereas **shortening the window** retroactively deletes older history org-wide and cannot be undone. Project-level retention overrides are no longer implied.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit df4f7551ed8c8e04e79b7e06d1d9d12cf61e5692. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/en/operate/governance/tool-executions/page.mdx
+++ b/app/en/operate/governance/tool-executions/page.mdx
@@ -3,9 +3,11 @@ title: "Tool Executions"
 description: "Review every tool run in a project: what ran, for whom, and why it failed, with tool inputs and outputs held to project admins."
 ---
 
+import { Callout } from "nextra/components";
+
 # Tool executions
 
-Arcade records every tool run in a project, including which tool ran, for which end user, when, and how it went, and exposes that history in the dashboard and over the API.
+Arcade records every tool run in a project, including which tool ran, for which end user, when, and how it went, and exposes that history in the dashboard and over the API. Recording is on by default and is an organization-wide setting — see [Recording and retention](#recording-and-retention).
 
 This page serves two jobs:
 
@@ -166,11 +168,50 @@ Two behaviors trip up API callers:
 
 Find your organization and project IDs in the dashboard URL: `/orgs/{org_id}/projects/{project_id}`. For the full request and response schemas, see the [API reference](/references/api).
 
-## Retention
+## Recording and retention
 
-Execution logging runs by default, and each project keeps its history for a bounded retention window: 7 days unless your organization sets a different default or a project overrides it. Past the window, Arcade deletes the records and they stop appearing in the history and the API.
+Two settings govern the history, both organization-wide and both set by an organization admin. They apply to every project in the organization.
 
-To change the retention window, or to turn off execution logging and delete the history for your organization, contact Arcade support.
+| Setting | What it decides |
+| -- | -- |
+| Recording | Whether new tool runs are written down at all |
+| Retention window | How long a recorded run is kept |
+
+Recording is on and the window is 7 days unless your organization changes them. Past the window, Arcade deletes the records and they stop appearing in the history and the API.
+
+### Change them yourself
+
+Open your organization in the dashboard and select **Execution Logging**, or call the API:
+
+```bash
+curl -s -X PUT "https://api.arcade.dev/v1/orgs/{org_id}/logging-config" \
+  -H "Authorization: Bearer $ARCADE_API_KEY" \
+  -H "Content-Type: application/json" \
+  -d '{"logging_data_retention": "ALLOWED", "default_log_retention_days": 30}'
+```
+
+Send either field on its own; an omitted field is left alone, so changing the window cannot switch recording off by accident.
+
+A window must be at least 1 day and no longer than the platform maximum, which is 90 days on Arcade Cloud and is reported as `max_log_retention_days` on the organization read. Anything outside that is refused with a `422` rather than quietly adjusted. Zero is not "keep nothing" — it is refused, because turning recording off is the way to stop recording.
+
+Changing the policy takes organization-admin authority. The organization read reports `can_edit_logging` so a client can tell whether the caller may change it before offering the control.
+
+### What each setting does to history you already have
+
+These come apart, and the difference matters:
+
+- **Turning recording off** stops new runs being written down. It leaves what is already stored in place, and those records go on expiring under the window as before. It is not an erasure control.
+- **Shortening the window** applies to records you already have. Going from 90 days to 7 deletes everything older than 7 days across every project in the organization, shortly after you save, and it cannot be undone.
+
+If you want existing history gone, shortening the window is the control that deletes it.
+
+<Callout type="info">
+Turning recording on permits projects to record; it does not compel it. While recording is organization-wide the two are indistinguishable, and that stops being true if project-level settings arrive later.
+</Callout>
+
+### When the history is empty
+
+An empty list has more than one cause, so the dashboard says which applies rather than leaving you to guess. If recording is off, Tool Executions says so instead of telling you that runs will appear once tools run. The project read also reports where the window in force came from and whether the project may change it, so a reader who cannot change it knows who to ask.
 
 ## Related content
 
--- a/app/en/operate/governance/tool-executions/page.mdx
+++ b/app/en/operate/governance/tool-executions/page.mdx
@@ -175,7 +175,7 @@ Two organization-wide settings govern the history, and are configured by an orga
 | Recording (boolean) | Whether new tool runs are written down at all |
 | Retention window (number) | How long a recorded run is kept |
 
-By default, recording is enabled and the retention window is 7 days. Past the window, Arcade deletes the records and they stop appearing in the history and the API.
+On Arcade Cloud, recording is enabled by default and the retention window is 7 days. Past the window, Arcade deletes the records and they stop appearing in the history and the API. A self-hosted deployment sets its own defaults, and records nothing until execution logging is turned on in the Engine configuration.
 
 ### Change them yourself
 
--- a/app/en/operate/governance/tool-executions/page.mdx
+++ b/app/en/operate/governance/tool-executions/page.mdx
@@ -3,8 +3,6 @@ title: "Tool Executions"
 description: "Review every tool run in a project: what ran, for whom, and why it failed, with tool inputs and outputs held to project admins."
 ---
 
-import { Callout } from "nextra/components";
-
 # Tool executions
 
 Arcade records every tool run in a project, including which tool ran, for which end user, when, and how it went, and exposes that history in the dashboard and over the API. Recording is on by default and is an organization-wide setting — see [Recording and retention](#recording-and-retention).
@@ -170,14 +168,14 @@ Find your organization and project IDs in the dashboard URL: `/orgs/{org_id}/pro
 
 ## Recording and retention
 
-Two settings govern the history, both organization-wide and both set by an organization admin. They apply to every project in the organization.
+Two organization-wide settings govern the history, and are configured by an organization admin. They apply to every project in the organization.
 
 | Setting | What it decides |
 | -- | -- |
-| Recording | Whether new tool runs are written down at all |
-| Retention window | How long a recorded run is kept |
+| Recording (boolean) | Whether new tool runs are written down at all |
+| Retention window (number) | How long a recorded run is kept |
 
-Recording is on and the window is 7 days unless your organization changes them. Past the window, Arcade deletes the records and they stop appearing in the history and the API.
+By default, recording is enabled and the retention window is 7 days. Past the window, Arcade deletes the records and they stop appearing in the history and the API.
 
 ### Change them yourself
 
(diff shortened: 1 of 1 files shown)


app/en/operate/governance/tool-executions/page.mdx
```diff
@@ -166,11 +168,50 @@ Two behaviors trip up API callers:
 
 Find your organization and project IDs in the dashboard URL: `/orgs/{org_id}/projects/{project_id}`. For the full request and response schemas, see the [API reference](/references/api).
 
-## Retention
+## Recording and retention
+
+Two settings govern the history, both organization-wide and both set by an organization admin. They apply to every project in the organization.
```
evantahler: This is very claude-y.  How about:<br><br>`Two organization-wide settings govern the history, and are configured by organization admin. They apply to every project in the organization.`

app/en/operate/governance/tool-executions/page.mdx
```diff
@@ -166,11 +168,50 @@ Two behaviors trip up API callers:
...
+
+| Setting | What it decides |
+| -- | -- |
+| Recording | Whether new tool runs are written down at all |
```
evantahler:
```
| Recording (boolean) | Whether new tool runs are written down at all |
| Retention window (number) | How long a recorded run is kept |
```

app/en/operate/governance/tool-executions/page.mdx
```diff
@@ -166,11 +168,50 @@ Two behaviors trip up API callers:
...
+| Retention window | How long a recorded run is kept |
+
+Recording is on and the window is 7 days unless your organization changes them. Past the window, Arcade deletes the records and they stop appearing in the history and the API.
```
evantahler: `By default, recording is enabled and the retention window is 7 days. Past the window, Arcade deletes the records and they stop appearing in the history and the API.`

app/en/operate/governance/tool-executions/page.mdx
```diff
@@ -166,11 +168,50 @@ Two behaviors trip up API callers:
...
+  -H "Authorization: Bearer $ARCADE_API_KEY" \
+  -H "Content-Type: application/json" \
+  -d '{"logging_data_retention": "ALLOWED", "default_log_retention_days": 30}'
+```
+
+Send either field on its own; an omitted field is left alone, so changing the window cannot switch recording off by accident.
+
+A window must be at least 1 day and no longer than the platform maximum, which is 90 days on Arcade Cloud and is reported as `max_log_retention_days` on the organization read. Anything outside that is refused with a `422` rather than quietly adjusted. Zero is not "keep nothing" — it is refused, because turning recording off is the way to stop recording.
+
+Changing the policy takes organization-admin authority. The organization read reports `can_edit_logging` so a client can tell whether the caller may change it before offering the control.
```
evantahler: This line is covered elsewhere - can be dropped

app/en/operate/governance/tool-executions/page.mdx
```diff
@@ -166,11 +168,50 @@ Two behaviors trip up API callers:
...
+
+### What each setting does to history you already have
+
+These come apart, and the difference matters:
```
evantahler: line can be removed

app/en/operate/governance/tool-executions/page.mdx
```diff
@@ -166,11 +168,50 @@ Two behaviors trip up API callers:
...
+
+- **Turning recording off** stops new runs being written down. It leaves what is already stored in place, and those records go on expiring under the window as before. It is not an erasure control.
+- **Shortening the window** applies to records you already have. Going from 90 days to 7 deletes everything older than 7 days across every project in the organization, shortly after you save, and it cannot be undone.
+
+If you want existing history gone, shortening the window is the control that deletes it.
```
evantahler: line can be removed

app/en/operate/governance/tool-executions/page.mdx
```diff
@@ -166,11 +168,50 @@ Two behaviors trip up API callers:
...
+
+<Callout type="info">
+Turning recording on permits projects to record; it does not compel it. While recording is organization-wide the two are indistinguishable, and that stops being true if project-level settings arrive later.
+</Callout>
```
evantahler: I don't think we have project-level settings yet - remove this

app/en/operate/governance/tool-executions/page.mdx
```diff
@@ -166,11 +168,50 @@ Two behaviors trip up API callers:
...
 
-Execution logging runs by default, and each project keeps its history for a bounded retention window: 7 days unless your organization sets a different default or a project overrides it. Past the window, Arcade deletes the records and they stop appearing in the history and the API.
+### When the history is empty
 
-To change the retention window, or to turn off execution logging and delete the history for your organization, contact Arcade support.
+An empty list has more than one cause, so the dashboard says which applies rather than leaving you to guess. If recording is off, Tool Executions says so instead of telling you that runs will appear once tools run. The project read also reports where the window in force came from and whether the project may change it, so a reader who cannot change it knows who to ask.
```
evantahler: I don't think this is nessicary - the UI explains what's going on

evantahler (changes requested):

app/en/operate/governance/tool-executions/page.mdx
josebarrueta: Taken, with one liberty: I added the article — "configured by **an** organization admin". Shout if you meant it without.

app/en/operate/governance/tool-executions/page.mdx
josebarrueta: Done.

app/en/operate/governance/tool-executions/page.mdx
josebarrueta: Taken verbatim.

app/en/operate/governance/tool-executions/page.mdx
josebarrueta: Dropped — your opening line already says an organization admin configures these, so the sentence was restating it.

app/en/operate/governance/tool-executions/page.mdx
josebarrueta: Gone. It announced a contrast the two bullets make on their own.

app/en/operate/governance/tool-executions/page.mdx
josebarrueta: Gone.

app/en/operate/governance/tool-executions/page.mdx
josebarrueta: Right, and the callout half-admitted it — "the two are indistinguishable" until project settings exist. Documenting a distinction with no observable effect is just something for a reader to trip over. Whole callout removed, and its `import { Callout }` with it.

app/en/operate/governance/tool-executions/page.mdx
josebarrueta: Fair — narrating UI copy in docs means two places to keep in step, and the UI wins that race. Section removed.

Worth knowing why it was there, so the intent is not lost: an empty list has several causes (nothing ran, recording off, window expired), and telling them apart is the outcome the feature is measured on. That job now sits entirely with the dashboard, which is the right place for it.

josebarrueta:
All eight applied — pushed as `10d24ff3`. Net **+4 / −20**, so the review took considerably more out than it put in, which was the right direction.

Your three wordings are in as written, bar one liberty: I added the article in "configured by **an** organization admin". Say the word if you meant it without.

The four cuts, and what each was doing:

- **Org-admin authority line** — your opening sentence already says it.
- **"These come apart, and the difference matters"** and **"If you want existing history gone…"** — both announced or restated the two bullets between them.
- **The permits-not-compels callout** — you're right that project-level settings don't exist, and the callout half-conceded it ("the two are indistinguishable"). A distinction with no observable effect is only something to trip over. Its `import { Callout }` went too, since that was the last use in the file.
- **"When the history is empty"** — narrating UI copy means two places to keep in step, and the UI wins that race.

One thing worth leaving on the record from that last cut: an empty list has several causes — nothing ran, recording is off, the window already expired — and telling them apart is the outcome this feature is measured on. That job now sits entirely with the dashboard, which is the better home for it, but it is a real requirement rather than a nicety.

`bun run build` passes.

app/en/operate/governance/tool-executions/page.mdx
josebarrueta: Following up on this one — I checked the sentence against the code and it needed scoping, so `df4f7551` amends what I took from you here.

**"recording is enabled" is Cloud-only.** The Engine has its own switch, and it defaults *off*:

```go
// internal/executionlogging/config.go
Enabled: false
```

`internal/config/config.go` builds from that default, and `engine.go` gates both the logger and the retention worker on it. So a self-hosted deployment that has not set `execution_logging.enabled: true` records nothing and runs no cleanup, whatever the organization policy says. On Cloud the sentence is right — the org starts `ALLOWED` and projects start `logging_enabled = true`.

Since the paragraph two below already says "90 days on Arcade Cloud", scoping this the same way makes the section read consistently:

> On Arcade Cloud, recording is enabled by default and the retention window is 7 days. […] A self-hosted deployment sets its own defaults, and records nothing until execution logging is turned on in the Engine configuration.

Two things I checked and left alone:

- **7 days** is `ORG_DEFAULT_LOG_RETENTION_DAYS`, deployment-overridable — covered by the same Cloud scoping.
- **"deletes … and they stop appearing"** is right in substance, slightly reversed in order. Records stop appearing when they are *marked* (every list, detail, and summary query filters `marked_for_deletion = FALSE`), and the sweep deletes them seconds later. Marking runs every 30 minutes by default, so a record can outlive its window by up to that long. Precise enough as written; the interval is deployment config and not worth a sentence.

Build passes.

evantahler (approved):

PR #1175, opened 2026-09-03
[GRO-303] DO NOT MERGE ME docs(agent-plugin): add install guides for the Arcade agent plugin

teallarson:
Closes [GRO-303](https://linear.app/arcadedev/issue/GRO-303/add-arcade-plugin-install-guides-to-docsarcadedev).

Adds a Get Started section for installing Arcade through the agent plugin, sourced from `docs/install/*` and `docs/support-matrix.md` in the [arcade-plugin](https://github.com/ArcadeAI/arcade-plugin) repo.

## Pages

`app/en/get-started/agent-plugin/`, placed second in Get Started because it's the only path where nothing has to be set up before you install. An Arcade account is still required; the difference is that the link your agent hands you on its first app task creates it, so signup happens at the moment of use rather than up front.

- **Overview** — the one-command install, `--target` variants, dry run, then per-agent verification in tabs (Cursor, Claude Code, VS Code, Copilot CLI, Codex), the support matrix, sign-in, and sample prompts
- **Claude Desktop** — connector URL, plus the Chat vs. Cowork/Code split
- **Any MCP client** — gateway URL for OpenCode and anything else speaking Streamable HTTP
- **From trial to team rollout** — your own gateway, a curated tool list, your own end-user identities, then the API and framework path

Tabs rather than a page per client: the per-client content is about eight lines each, and five near-identical pages would rot independently.

The existing MCP Clients guides all assume you want your own gateway and an API key, so each one now has a callout pointing trial users at the plugin, and the plugin pages link back for anyone who needs their own tool list and identities.

## The URL is a variable

`lib/agent-plugin.ts` holds two constants:

```ts
export const AGENT_PLUGIN_GATEWAY_URL = "https://api.bosslevel.dev/mcp/all-optimized";
export const AGENT_PLUGIN_REPO = "ArcadeAI/arcade-plugin";
```

Everything else derives from those — the install commands, and the Cursor and VS Code one-click links, whose base64 and percent-encoded configs are built at compile time rather than pasted. The generated Cursor link came out byte-identical to the hand-written one in the plugin repo.

Pages reference them as `{{ARCADE_PLUGIN_GATEWAY_URL}}` and friends. `lib/remark-substitute.ts` resolves those in prose, inline code, fenced code blocks, and link targets. Unknown tokens pass through, so the `{{client_id}}` placeholders on the auth-provider pages still work.

Two notes for whoever maintains this:

- The plugin also rewrites the raw source string, not just the AST. Nextra exports that separately and it feeds the copy-page button and the markdown this site serves to agents. Without it, agents read the literal token while humans read the URL.
- `pnpm build` caches compiled MDX. After changing a constant, `rm -rf .next` or the old value survives in the output.

`tests/broken-link-check.test.ts` now resolves tokens before validating, otherwise it reads a token as a relative path.

## Not included

- **The `.mcpb` for Claude Desktop.** No such artifact exists in the plugin repo yet (GRO-302), so the page documents the connector URL only.
- **The `mcp-remote` sample config**, which points at a file inside the private repo.
- **A link to the plugin repo**, which would 404 while it's private.

## Why this is a draft

`npx plugins add ArcadeAI/arcade-plugin` does not work for anyone outside the org until the repo goes public, and the gateway constant is still staging. Both are gated on GRO-309. Swap that one line and undraft when the production endpoint and marketing page are ready.

## Checks

`pnpm build`, `pnpm lint`, and all 881 tests pass. Vale reports `0 errors, 0 warnings, 0 suggestions` on the six touched pages it can parse.

Vale can't parse the other four (`mcp-clients/cursor`, `claude-code`, `claude-desktop`, `github-copilot`), and neither can it parse 33 of the repo's 178 pages. The MDX is valid — `mdx2vast` reads all of them raw. The trigger is the `TokenIgnores` backtick pattern in `.vale.ini`, which blanks inline-code spans and unbalances the JSX. Since Vale aborts the whole run on the first crash, `pnpm vale:check` never completes. Separate PR; I checked those four files by hand.


app/en/get-started/agent-plugin/tools-only/any-mcp-client/page.mdx
```diff
@@ -0,0 +1,44 @@
...
+{{ARCADE_PLUGIN_GATEWAY_URL}}
+```
+
+Your agent gets every tool on the gateway. Nothing else about how you use it changes.
+
+## One-click links
+
+Cursor and VS Code can add the gateway without editing any config:
+
+- **Cursor:** use [this link]({{ARCADE_PLUGIN_CURSOR_INSTALL_LINK}}) for a one-click MCP install.
```
torresmateo, line 20: non-blocking: perhaps we could reuse the component from the dashboard here with the client dropdown and only show the clients that support plugins?

app/en/get-started/agent-plugin/page.mdx
```diff
@@ -0,0 +1,127 @@
...
+
+To see what an install would write without changing anything:
+
+```bash
+npx plugins discover {{ARCADE_PLUGIN_REPO}}
+```
+
+Reload your agent afterwards if it does not pick up the plugin on its own. A couple of agents gate plugins behind a setting. Claude Code may ask you to enable it with `claude plugin`, and VS Code needs `chat.plugins.enabled` set to `true`.
+
+Two things worth knowing if you use more than one agent. Codex and ChatGPT share a plugin directory, so one install covers both. And VS Code can discover a copy installed through GitHub Copilot CLI, so install in one place rather than both.
```
torresmateo, line 62:
```suggestion
If you use more than one agent, you can install the plugin in one place. Codex and ChatGPT share a plugin directory, and VS Code can discover a copy installed through GItHub Copilot CLI.
```

(not sure if some meaning is lost in my version, feel free to ignore

app/en/get-started/agent-plugin/page.mdx
```diff
@@ -0,0 +1,127 @@
...
+
+// todo: claude mcpb instructions
+
+Using a client that does not load plugins? Those connect to the same gateway as a [tools-only MCP server](/get-started/agent-plugin/tools-only).
```
torresmateo, line 66: nit: should we call it "tools-only Arcade bundle" here for consistency?

app/en/get-started/agent-plugin/page.mdx
```diff
@@ -0,0 +1,127 @@
...
+<GuideOverview.Outcomes>
+
+Install the Arcade agent plugin in your agent and run your first task against a connected app.
+
+</GuideOverview.Outcomes>
+
+<GuideOverview.Prerequisites>
+
+1. [Node.js](https://nodejs.org) so you can run `npx`
+2. A supported agent: Cursor, Claude Code, VS Code, GitHub Copilot CLI, or Codex
```
torresmateo:
```suggestion
2. An agent with [plugins](https://agent-plugins.org) support: Cursor, Claude Code, VS Code, GitHub Copilot CLI, or Codex
```

app/en/get-started/agent-plugin/page.mdx
```diff
@@ -0,0 +1,127 @@
...
+| Agent                |     Tools      | Skills | Subagent | Commands | Rule | Hooks |
+| -------------------- | :------------: | :----: | :------: | :------: | :--: | :---: | --- | -------------- |
+| Cursor               |       ✅       |   2    |    ✅    |    3     |  ✅  |  ✅   |
+| Claude Code          |       ✅       |   2    |    ✅    |    3     |  —   |  ✅   |
+| Claude Cowork        |       ✅       |   2    |    ✅    |    3     |  —   |  ✅   |
+| GitHub Copilot CLI   |       ✅       |   2    |    ✅    |    —     |  —   |   —   |
+| VS Code              |       ✅       |   2    |    —     |    —     |  —   |   —   |
+| Codex and ChatGPT    |       ✅       |   2    |    —     |    —     |  —   |   —   |
+| Any other MCP client |       ✅       |   —    |    —     |    —     |  —   |   —   |
+| //                   | Claude Desktop |   ✅   |    —     |    —     |  —   |   —   | —   | TODO: not true |
```
torresmateo, line 94: nit: this table renders as text on the preview

app/en/get-started/agent-plugin/page.mdx
```diff
@@ -0,0 +1,127 @@
...
+
+Here is what each of those is for.
+
+**Commands** are the part you type. `/arcade-status` reports whether the gateway is reachable, whether you have signed in, and which apps you have authorized. `/arcade-connect <app>` authorizes an app on demand, so you are not interrupted mid-task. `/arcade-apps` lists what you have connected and lets you disconnect one.
```
torresmateo, line 98: suggestion: use the style we use on the CLI reference for these commands: https://docs-git-teal-gro-303-add-arcade-plugin-instal-8ebd0e-arcade-ai.vercel.app/en/references/arcade-cli

app/en/get-started/agent-plugin/tools-only/page.mdx
```diff
@@ -0,0 +1,43 @@
...
+
+## What you give up
+
+Tools alone still do the work. What you lose is the part that helps your agent use them well:
+
+- **The skills** teach your agent to turn "reply to that thread" into the right sequence of tool calls, and to walk you through a team rollout when you ask about one.
+- **The commands** give you `/arcade-status`, `/arcade-connect`, and `/arcade-apps` for checking and managing connections without leaving your agent.
+- **The subagent** keeps tool search out of your main conversation, so it does not crowd out your actual work.
+
+If your client supports plugins, [install the full plugin](/get-started/agent-plugin) instead.
```
torresmateo:
```suggestion
If your client supports [plugins](https://agent-plugins.org), you should [install the full plugin](/get-started/agent-plugin) instead.
```

torresmateo (changes requested):
I like these docs with a small caveat.

I think we're introducing two things that look very similar to MCP gateways and if you're not deep in the weeds of what's a plugin vs a gateway vs a server things can spiral very quickly. 

On the overview page, I think we need a sentence on the plugin overview page that explains what it is as clearly as possible, telling users the things that are bundled at the beginning of that page.

app/en/get-started/agent-plugin/page.mdx
teallarson: I think this sentence can just go away.

app/en/get-started/agent-plugin/page.mdx
teallarson: TY! I put a TODO in the middle of hte table and broke it 🤦‍♀️

PR #1185, opened 2026-09-08
docs(contextual-access): document group-based tool access

andrewconnor-arcade:
Public configuration page for the `group_access` contextual-access extension, alongside the existing [rate limiting](https://docs.arcade.dev/en/operate/governance/contextual-access/rate-limiting) page and following its structure.

**Draft until the engine work merges.** The feature ships as a stack in the monorepo (PLT-3229 through PLT-3234); this page describes behavior that is built and green but not yet on `main`. Holding it in draft so the docs never describe something a customer cannot use yet.

## What it covers

Enforcement at the tool access hook point and why it governs both discovery and execution, tool matcher precedence, the three identity backends and what access each needs, Dashboard and API setup, the staleness model, and fail-closed behavior with the per-rule opt-out.

## Two things given deliberate weight

Both fail *silently* rather than loudly, which is what makes them worth a customer's attention up front:

- **Identity matching.** The backend looks a caller up by the identity the request carries, so an Arcade user ID that does not equal the directory login matches nothing, and fail-closed behavior turns that into tools quietly missing rather than an error anyone can act on.
- **SailPoint's API client rights.** `idn:access-profiles:read` and `idn:roles:read` sound like the right rights and are not sufficient: they authorize reading access profile and role *definitions*, not which identity holds them. A client granted only those authenticates successfully and then fails every membership lookup. This is the single most likely misconfiguration, so it gets a warning callout rather than a footnote.

The page also states that **a tool no rule matches stays visible**, so a mistyped matcher governs nothing instead of hiding everything, and recommends gating on the provisioned Okta or Entra group where SailPoint provisions into one, since that is simpler to operate and closer to real time.

## Checks

Vale is clean of errors on the new page. The remaining warnings are Microsoft's own field names (`application (client) ID`, `GroupMember.Read.All` application permission) and heading-case rules the sibling rate-limiting page also trips; I matched the section's existing conventions rather than diverging from its neighbor. Pre-commit hooks (Vale, `_meta.tsx` key validation, redirect check, Ultracite) passed.

Refs PLT-3234.

---

> [!NOTE]
> **Low Risk**
> Documentation-only changes; no application or engine code in this diff.
> 
> **Overview**
> Adds a **Group-Based Tool Access** doc under contextual access, wired into the section nav via `_meta.tsx` and listed in `public/llms.txt`.
> 
> The new page documents the native `group_access` extension: enforcement at the tool access hook (discovery and execution), tool matcher precedence, Okta / Entra / SailPoint backends and credentials, Dashboard and plugins API setup, cache staleness, and fail-closed behavior with per-rule `allow_on_unavailable`. It highlights operational pitfalls—Arcade user ID must match directory login, Entra object IDs for sensitive rules, SailPoint API scopes that look correct but are insufficient, and that unmatched tools stay visible.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f1e4853a0c64914d96f0c88d0fe200afac5b12a5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/en/operate/governance/contextual-access/group-access/page.mdx
+++ b/app/en/operate/governance/contextual-access/group-access/page.mdx
@@ -1,0 +1,203 @@
+---
+title: "Group-Based Tool Access"
+description: "Vary which tools a caller can see and use based on their identity provider group membership, enforced natively by the Arcade Engine"
+---
+
+import { Callout, Steps } from "nextra/components";
+
+# Group-Based Tool Access
+
+As a platform operator, you can vary which tools a caller sees based on who they are. Group-based access is built into the Arcade Engine: you define rules that require membership in an identity provider group, and the Engine resolves membership and filters the tool list before the caller ever sees it. Like rate limits, this runs natively inside Arcade, so there is no server to build or host.
+
+Use it when one shared MCP gateway serves more than one team or customer and they should not all see the same tools. Without it, the alternatives are a separate gateway per access boundary or a custom webhook extension.
+
+## How enforcement works
+
+A group access configuration has two parts: an **identity backend** that holds your groups, and a set of **rules**. Each rule combines two things:
+
+- A **tool matcher** that selects which tools the rule applies to
+- One or more **required groups** the caller must belong to
+
+Enforcement runs at the tool access hook point, so it applies to both discovery and execution:
+
+- A caller who does not satisfy a matched rule has the tool **removed from the response** rather than being shown a tool that returns an error.
+- Naming that tool directly is also refused, so hiding is not the only enforcement. A caller who guesses a tool name gains nothing.
+
+Membership is resolved by querying your identity provider, then cached per user and per group for five minutes by default.
+
+<Callout type="info">
+  Multiple required groups on one rule are combined with OR: the caller needs
+  membership in any one of them, not all of them.
+</Callout>
+
+### Tool matchers
+
+| Matcher | Example | Applies to |
+| --- | --- | --- |
+| Exact | `Slack.SendMessage` | One fully qualified tool |
+| Toolkit | `Slack.*` | Every tool in the toolkit |
+| Global | `*` | Every tool |
+
+When several rules match the same call, only the most specific rule applies: an exact match beats a toolkit match, and a toolkit match beats the global match. The caller is evaluated against that one rule only.
+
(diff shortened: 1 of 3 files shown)


app/en/operate/governance/contextual-access/group-access/page.mdx
evantahler: This content is great, but it's pretty "claude-y".  Can you give it a tone/voice review please?

app/en/operate/governance/contextual-access/group-access/page.mdx
```diff
@@ -0,0 +1,255 @@
...
+
+### Tool matchers
+
+| Matcher | Example | Applies to |
+| --- | --- | --- |
+| Exact | `Slack.SendMessage` | One fully qualified tool |
+| Toolkit | `Slack.*` | Every tool in the toolkit |
+| Global | `*` | Every tool |
+
+Group names are matched **exactly, including case**, while tool matchers are not. A rule requiring `Finance-Admins` is not satisfied by membership in `finance-admins`. The asymmetry is deliberate: a mis-cased tool matcher governs nothing, whereas a loosely matched group would widen who satisfies the rule. Group IDs are matched without regard to case, since an ID is opaque.
```
evantahler: This really make me want a drop-down list of possible groups to avoid this.  That's phase 2 I guess!

evantahler (approved):
Approving generally, because this is OK to ship as-is, but please take a pass on removing the claude-isms

app/en/operate/governance/contextual-access/group-access/page.mdx
andrewconnor-arcade: That's PLT-3662! Definitely want to get it for the GA release

app/en/operate/governance/contextual-access/group-access/page.mdx
andrewconnor-arcade: did a tone pass in 3c326299 to match the rate-limiting page: cut about a quarter, dropped the rationale asides and internal detail. also fixed a few things along the way: okta and entra permissions are now documented, and a wrong claim that the cache TTL is configurable is gone (PLT-3702 tracks making it configurable)

